PR #60, opened 2016-09-25
AppVeyor integration

tusharpm:
CI integration for Windows, mostly on the same lines as travis-ci integration. Building with MSVC however, needed non-trivial changes in the C++ code because of lack of standard compliance.
Here is a summary of the changes:
- In `enum class UVStdIOFlags`, changed `IGNORE` to `IO_IGNORE` because the former is a macro in a Windows header.
- `StreamHandle` exposes only one inherited constructor from `Handle` on Windows. This, for now, is also the only one but will need some work when adding another.
- `Flags`' assignment operators are not `constexpr` on Windows (as explained in a comment). This can affect dependent functions.
- Renamed all instances of the token `interface` as it is a macro in a Windows header.

Other changes (outside C++ code):
- CMake module file changes to pick the dependencies from the correct build location
 - This includes one change to reduce warnings on Windows (from around 16k to 60) reducing the time taken for build completion (from ~28 minutes to ~2 minutes) per platform.
- Batch script equivalents of
 - `deps.sh` -> `deps.bat`
 - one script to build and run tests -> `runtests.bat`
- AppVeyor configuration file

Potential changes which were left out:
- Instructions to run tests on a local machine
- The AppVeyor badge
- Email notifications from AppVeyor

I trust Travis to ensure nothing else broke on Linux/macOS because of these changes. Please let me know if there's something to be done manually. Also, let me know if this needs some changes.


cmake/modules/FindUV.cmake
```diff
@@ -22,14 +22,16 @@ find_path(
 )
 
 find_library(
-    UV_STATIC_LIBRARY NAMES libuv.a
-    PATHS ${BUILD_DEPS_DIR}/${UV_DEPS_DIR}/.libs/
+    UV_STATIC_LIBRARY NAMES libuv
```
skypjack: Isn't the  `.so` object picked up if `.a` is not specified?

src/uvw/process.hpp
```diff
@@ -28,7 +28,7 @@ enum class UVProcessFlags: std::underlying_type_t<uv_process_flags> {
 
 
 enum class UVStdIOFlags: std::underlying_type_t<uv_stdio_flags> {
-    IGNORE = UV_IGNORE,
+    IO_IGNORE = UV_IGNORE,
```
skypjack: Documentation must be updated as well (see the `stdio` member methods of the class `ProcessHandle` in the same file).

src/uvw/stream.hpp
```diff
@@ -171,7 +171,9 @@ class StreamHandle: public Handle<T, U> {
     }
 
 protected:
-    using Handle<T, U>::Handle;
+	StreamHandle(std::shared_ptr<Loop> ref) noexcept
```
skypjack: `noexcept` should depend on the underlying constructor, otherwise `std::terminate` will be called in case of exceptions.

src/uvw/util.hpp
```diff
@@ -314,7 +320,7 @@ Addr address(F &&f, const H *handle) noexcept {
 template<typename F, typename... Args>
 std::string path(F &&f, Args... args) noexcept {
     std::size_t size = DEFAULT_SIZE;
-    char buf[size];
+    char buf[DEFAULT_SIZE];
```
skypjack: Within this method, use `size` everywhere for consistency. Do not use `DEFAULT_SIZE` to declare the buffer. Thank you.

src/uvw/udp.hpp
```diff
@@ -246,13 +246,13 @@ class UDPHandle final: public Handle<UDPHandle, uv_udp_t> {
      * * `UDPHandle::Membership::JOIN_GROUP`
      *
      * @param multicast Multicast address to set membership for.
-     * @param interface Interface address.
+     * @param interface_addr Interface address.
```
skypjack: `uvw` coding style comprises camel case, `libuv` uses underscores. I would suggest `iface` instead of `interface_addr`.

src/uvw/udp.hpp
```diff
@@ -278,12 +278,12 @@ class UDPHandle final: public Handle<UDPHandle, uv_udp_t> {
 
     /**
      * @brief Sets the multicast interface to send or receive data on.
-     * @param interface Interface address.
+     * @param interface_addr Interface address.
```
skypjack: `uvw` coding style comprises camel case, `libuv` uses underscores. I would suggest `iface` instead of `interface_addr`.

src/uvw/util.hpp
```diff
@@ -11,6 +11,12 @@
 #include <array>
 #include <uv.h>
 
+#ifdef _WIN32
```
skypjack: I suspect `R_CONSTEXPR` is useless, even if MSVC has not relaxed constexpr's contraints. See the comments below for further details.

src/uvw/util.hpp
```diff
@@ -97,12 +103,12 @@ class Flags final {
 
     ~Flags() noexcept { static_assert(std::is_enum<E>::value, "!"); }
 
-    constexpr Flags& operator=(const Flags &f) noexcept {
+    R_CONSTEXPR Flags& operator=(const Flags &f) noexcept {
```
skypjack: You can define this by means of the comma operator and work around the constraints on the `constexpr` for MSVC.  
It follows an example:
```
constexpr Flags& operator=(const Flags &f) noexcept {
    return flags = f.flags, *this;
}
```

src/uvw/util.hpp
```diff
@@ -97,12 +103,12 @@ class Flags final {
...
         flags = f.flags;
         return *this;
     }
 
-    constexpr Flags& operator=(Flags &&f) noexcept {
+    R_CONSTEXPR Flags& operator=(Flags &&f) noexcept {
```
skypjack: You can define this by means of the comma operator and work around the constraints on the `constexpr` for MSVC.  
It follows an example:
```
constexpr Flags& operator=(Flags &&f) noexcept {
    return flags = std::move(f.flags), *this;
}
```

src/uvw/util.hpp
```diff
@@ -507,21 +513,21 @@ struct Utilities {
 
         if(0 == uv_interface_addresses(&ifaces, &count)) {
             std::for_each(ifaces, ifaces+count, [&interfaces](const auto &iface) {
-                InterfaceAddress interface;
+                InterfaceAddress tInterface;
```
skypjack: I would use `iface`. `tInterface` is somehow misleading. Thank you.

cmake/modules/FindUV.cmake
tusharpm: I thought it was the other way around.
The Travis build reports for the pull request show that they chose `libuv.a` to represent UV on both [precise](https://travis-ci.org/skypjack/uvw/jobs/162603519#L609) and [macOS](https://travis-ci.org/skypjack/uvw/jobs/162603520#L310).
Also, the tests passed afterwards, which wouldn't have on macOS due to lack of rpath configuration.

src/uvw/process.hpp
tusharpm: That makes sense. Sorry about that. I'll push these in the next commit.

src/uvw/stream.hpp
tusharpm: Nice catch. Will drop the `noexcept` specifier.

src/uvw/util.hpp
tusharpm: I get the following error on my machine with or without the workaround mentioned below:
```
t:\gitrepo\uvw\src\uvw\util.hpp(107): error C3490: 'flags' cannot be modified because it is being accessed through a const object [T:\GitRepo\uvw\build\test\odr.vcxproj]
t:\gitrepo\uvw\src\uvw\util.hpp(107): error C2440: 'return': cannot convert from 'const uvw::Flags<uvw::ProcessHandle::Process>' to 'uvw::Flags<uvw::ProcessHandle::Process> &' [T:\GitRepo\uvw\build\test\odr.vcxproj]
```
Where the lines 106-108 in `util.hpp` are the ones taken from the example.
I'm certain that the same will happen on AppVeyor machines as well. Please tell me if a test is needed.

src/uvw/util.hpp
tusharpm: Umm, `iface` is the name of a lambda parameter. Please suggest another name.
Also, I have a suggestion for this lambda (avoiding the need of a second name altogether): to use a constructor for `InterfaceAddress` taking a `const uv_interface_address_t &` (viz. `iface` here). We could benefit from using `std::vector<InterfaceAddress>::emplace_back`.

cmake/modules/FindUV.cmake
skypjack: Fair enough. I don't know why I was sure of the opposite. Nice catch.

src/uvw/process.hpp
skypjack: Changed to `IGNORE_STREAM`, it seems more appropriate. Pushed on the PR.

src/uvw/stream.hpp
skypjack: Why can't we use the inherited constructor here? I mean, we use them all around the code (see `Handle` itself), why doesn't this one work? Are we solving the wrong problem? It goes without saying that inherited constructors are accepted even on MSVC (otherwise you shouldn't be able to create any handle), so what?

src/uvw/util.hpp
skypjack: Is it a MSVC problem? The current code uses `char buf[size]` and compiles fine. Anyway, if it doesn't work, let's use `DEFAULT_SIZE` to initialize `size` and that's all. Much better than `std::size_t size = sizeof(buf);`. :-)

src/uvw/util.hpp
skypjack: They are aggregates, I'd rather exploit this. I'm modifying the commit on the fly, let me know if it works for you (it should).

src/uvw/util.hpp
skypjack: I can't reproduce the error. The changes below compile fine. If you agree, I would apply them and merge this PR, then we can look over it and decide how to proceed. How did you reproduce the error? What's the environment?

skypjack:
Currently, after the changes, my only concern are on `StreamHandle` and `R_CONSTEXPR`.

They seems _XY-problems_:
* I even don't know why a change to the `StreamHandle`'s constructor is required
* I cannot reproduce the error and the proposed changes work just fine with not relaxed `constexpr`s

For I cannot test/reproduce them, can you help me to solve them? I'm willing to merge the PR, of course.

Thank you.

src/uvw/util.hpp
tusharpm: Mine is a Window 10 Home x64 machine with Visual studio Community 2015 Update 3.
It is very close (except for the OS flavor of course) to the image used by AppVeyor.
Here are the test results from AppVeyor for [Win32](https://ci.appveyor.com/project/tusharpm/uvw/build/1.0.17/job/4khwdohx7372h5da/messages) and [x64](https://ci.appveyor.com/project/tusharpm/uvw/build/1.0.17/job/6f45r2rfh0lk52h9/messages) too, just for closure.

src/uvw/stream.hpp
tusharpm: Agreed: MSVC understands inherited constructors. The problem is that it gets confused easily. It complains that the constructor being inherited is ambiguous, most likely because of the multiple inheritance done by `Handle`. It seems to have no problem for other classes being compiled.
Please have a look at [the AppVeyor build report](https://ci.appveyor.com/project/tusharpm/uvw/build/job/9bklv1yv98xa6n5t/messages)([s](https://ci.appveyor.com/project/tusharpm/uvw/build/job/fi2kl0q32fraq6os/messages)) and let me know the next steps.

src/uvw/util.hpp
skypjack: Interesting. From the logs it seems that you are not using revision 14 of the standard (actually, it mentions the fact that with C++14 `const` won't be implied anymore), or at least you have not all the available features at your disposal. Anyway, good catch. Thank you for the help.

src/uvw/stream.hpp
skypjack: MSVC is somehow intriguing. :+1:

skypjack:
I'm going to merge the PR. Minor changes will follow on the master branch. Thank you for your help.

tusharpm:
Thanks.

tusharpm:
Hi Michele,

Please enable the AppVeyor integration in the project settings. It will
help you to easily verify if any incoming pull request breaks a Windows
build.
You will have to sign up for AppVeyor much the same way as Travis.

Thanks,
-T

skypjack:
Just done. Thank you.

PR #61, opened 2016-10-02
Cleanup

tusharpm:
This is an attempt to reduce the number of times `uv_*_t` identifiers are mentioned in uvw code.
Since they appear in most declarations (template parameter of the class/member function), the identifier itself can be removed in favor of a template parameter, which can further be left for the compiler to deduce.

This also reduces the repetition of `uv_*_t` in the same class. Using the declaration of the class template parameters, members can directly `get` the same inner uv type without using the name and specifying an independent template parameter (see `Resource::get`, `Handle::initialize` and `Request::defaultCallback`).

One seemingly unrelated change is in `lib.hpp` where a locally defined details::IsFunc is removed in favor of [the standard `std::is_function`](http://en.cppreference.com/w/cpp/types/is_function). This allows `F` to take more lenient forms like variadic functions, etc.

### DIFF
--- a/src/uvw/emitter.hpp
+++ b/src/uvw/emitter.hpp
@@ -177,7 +177,9 @@ public:
      */
     void clearAll() noexcept {
         for(auto &&h: handlers) {
-            h->clear();
+            if (h) {
+                h->clear();
+            }
         }
     }
 
@@ -192,7 +194,7 @@ public:
 
         return (!(type < handlers.size()) ||
                 !handlers[type] ||
-                static_cast<Handler<E>&>(*handlers[type]).empty());
+                handlers[type]->empty());
     }
 
     /**
--- a/src/uvw/emitter.hpp
+++ b/src/uvw/emitter.hpp
@@ -7,7 +7,6 @@
 #include <cstddef>
 #include <vector>
 #include <memory>
-#include <tuple>
 #include <list>
 #include "event.hpp"
 
--- a/src/uvw/emitter.hpp
+++ b/src/uvw/emitter.hpp
@@ -177,9 +177,7 @@ public:
      */
     void clearAll() noexcept {
         for(auto &&h: handlers) {
-            if (h) {
-                h->clear();
-            }
+            h->clear();
         }
     }
 
@@ -194,7 +192,7 @@ public:
 
         return (!(type < handlers.size()) ||
                 !handlers[type] ||
-                handlers[type]->empty());
+                static_cast<Handler<E>&>(*handlers[type]).empty());
     }
 
     /**
--- a/src/uvw/lib.hpp
+++ b/src/uvw/lib.hpp
@@ -4,21 +4,12 @@
 #include <utility>
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <uv.h>
 #include "loop.hpp"
 
 
 namespace uvw {
-
-
-namespace details {
-
-
-template<typename T> struct IsFunc: std::false_type { };
-template<typename R, typename... A> struct IsFunc<R(A...)>: std::true_type { };
-
-
-}
 
 
 /**
@@ -73,7 +64,7 @@ public:
      */
     template<typename F>
     F * sym(std::string name) {
-        static_assert(details::IsFunc<F>::value, "!");
+        static_assert(std::is_function<F>::value, "!");
         F *func;
         auto err = uv_dlsym(&lib, name.data(), reinterpret_cast<void**>(&func));
         if(err) { func = nullptr; }
--- a/src/uvw/lib.hpp
+++ b/src/uvw/lib.hpp
@@ -6,7 +6,6 @@
 #include <string>
 #include <uv.h>
 #include "loop.hpp"
-#include "loop.hpp"
 
 
 namespace uvw {
--- a/src/uvw/request.hpp
+++ b/src/uvw/request.hpp
@@ -22,8 +22,8 @@ protected:
         return ptr;
     }
 
-    template<typename R, typename E>
-    static void defaultCallback(R *req, int status) {
+    template<typename E>
+    static void defaultCallback(U *req, int status) {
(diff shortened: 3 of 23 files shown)


src/uvw/emitter.hpp
```diff
@@ -178,7 +177,9 @@ class Emitter {
      */
     void clearAll() noexcept {
         for(auto &&h: handlers) {
-            h->clear();
+            if (h) {
```
skypjack: This is a bug fix.
You should rather create two PR when you want to fix something and to propose changes.
This way we can speed up one PR and review the other one.

src/uvw/emitter.hpp
```diff
@@ -193,7 +194,7 @@ class Emitter {
 
         return (!(type < handlers.size()) ||
                 !handlers[type] ||
-                static_cast<Handler<E>&>(*handlers[type]).empty());
+                handlers[type]->empty());
```
skypjack: This is not an improvement.
We know the exact type of the handler (that is `E`) and casting it allows a direct call to the method in the final class.
On the other side, you are proposing to call a virtual method, thus to introduce a double dispatch that gives no benefits.
I'd suggest to rollback this change.

src/uvw/lib.hpp
```diff
@@ -74,7 +64,7 @@ class SharedLib final {
      */
     template<typename F>
     F * sym(std::string name) {
-        static_assert(details::IsFunc<F>::value, "!");
+        static_assert(std::is_function<F>::value, "!");
```
skypjack: Good catch. Shame on me. :+1:

src/uvw/request.hpp
```diff
@@ -22,8 +22,8 @@ class Request: public Resource<T, U> {
         return ptr;
     }
 
-    template<typename R, typename E>
-    static void defaultCallback(R *req, int status) {
+    template<typename E>
+    static void defaultCallback(U *req, int status) {
```
skypjack: That's good. I introduced the `U` parameter during the second phase of the project and I forgot to remove this stuff. Nice catch.

src/uvw/resource.hpp
```diff
@@ -29,19 +29,19 @@ class Resource: public Emitter<T>, public std::enable_shared_from_this<T> {
           pLoop{std::move(ref)},
           resource{}
     {
-        this->template get<U>()->data = static_cast<T*>(this);
+        resource.data = static_cast<T*>(this);
```
skypjack: It makes sense.
Initially `resource` was a _type-erased_ data member and I used to cast it.
Nowadays it is no longer required actually.

src/uvw/resource.hpp
```diff
@@ -29,19 +29,19 @@ class Resource: public Emitter<T>, public std::enable_shared_from_this<T> {
...
     }
 
     auto parent() const noexcept {
         return pLoop->loop.get();
     }
 
-    template<typename R>
+    template<typename R = U>
```
skypjack: Give me a while to think about it.
I don't like it too much, for we are _hiding_ the type during the call to `get` and I usually like this kind of things to be explicit.
Moreover, in terms of code bloating, there is no benefit, for you are not removing a template specialization.
Anyway, I see your point, even though I'm not sure it's really an improvement in terms of readability.

src/uvw/emitter.hpp
tusharpm: I'll revert this here, and open another PR for this change.

src/uvw/emitter.hpp
tusharpm: Cool, will rollback.

src/uvw/resource.hpp
tusharpm: My original intent was to add another protected member, something like `getNative` (one `const` qualified overload, another not) that returns the raw pointer `U*` (or `const U*` accordingly). That would avoid the template instantiation altogether. But it involves changing the used `get` function in a lot of places.
Please tell me if that is a favorable option.

skypjack:
Travis failed to compile it (clang). Is it something due to these changes?

src/uvw/emitter.hpp
skypjack: I'm doing a review of `emitter.hpp`, I'll add the patch to that commit. Thank you.

src/uvw/resource.hpp
skypjack: You don't have to pick a different name up at all. You can simply remove the default argument and add the following functions:
```
    auto get() noexcept {
        return &resource;
    }

    auto get() const noexcept {
        return static_cast<const U*>(&resource);
    }
```
Non template member methods will be chosen whenever you don't explicitly specify a template parameter, otherwise the template function will be used.
I guess this is probably the best solution to reduce both the boilerplate and the code bloating.

Could you proceed with the change?

Thank you very much.

tusharpm:
It didn't fail to build, I think. I've seen it happen on my builds sometimes. Usually it goes away with just a rebuild. But this also points that we should try to get the output of the tests at least in case of failures.
Please have a look at [how googletest does it](https://github.com/google/googletest/blob/master/travis.sh).

src/uvw/resource.hpp
tusharpm: I'll make the suggested changes.
I have a quick question though: do we need the `static_cast` in the `const` qualified non-template `get`? I think we could make do with just a `&resource` expression.
Ideas?

tusharpm: I'm not too sure about that. I tried playing around with some code to help me understand how it would work. It seems `auto` deduces pretty well for the considerably new feature it is.
[Here's a test program](http://coliru.stacked-crooked.com/a/611d106f1a945420) demonstrating the same.
I'm not sure if it is compiler specific, I can try running the same program in VS on my machine.
Also, if it comes to it, shouldn't we be choosing `const_cast<const U*>`?

PR #62, opened 2016-10-07
Code coverage using coveralls.io

tusharpm:
Added the code coverage reporting and integrated with [coveralls.io](https://coveralls.io/) for visualizing and tracking the coverage numbers. [Here](https://coveralls.io/builds/8234660) is a sample report from the test build.
<br />* Reorganized the Travis configuration to streamline adding new builds/tool-chains later on.
<br />* Added the new jobs in travis:
<br /> - `g++-6` for Linux
<br /> - `xcode8` for macOS
<br />
<br />Left the badge for `README.md` to be added later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ matrix:
     before_script:
     - pip install --user cpp-coveralls
     after_success:
-    - coveralls --gcov gcov-6 --gcov-options '\-lp' --exclude deps
+    - coveralls --gcov gcov-6 --gcov-options '\-lp' --root ${TRAVIS_BUILD_DIR} --build-root ${TRAVIS_BUILD_DIR}/build --exclude deps --include src
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


skypjack:
So, should I create an account on _coveralls.io_?
What kind of coverage is it reporting? Tests do not cover files like `thread.hpp` and some others, but it is saying that they are 100% covered.

tusharpm:
Yes, you'll have to sign-up on coveralls to use the integration.
For the reporting, I'm not sure why it reports files with 0 relevant/covered lines as 100% covered. Anyway, sorting by relevant lines (descending) brings the files that are actually covered by the test code to the top. The overall percentage reported is calculated as the ratio of total covered lines to total relevant lines. Essentially, until we write a test case to exercise a header, it doesn't contribute to the percentage shown in the reports.

skypjack:
Is there a way to have the actual coverage?
I mean, if a function or a file has not been visited, it should be reported as 0%.
Today it is excluded and doesn't influence the percentage at all.
I prefer a realistic 23% of coverage than an unrealistic 97%!!

PR #69, opened 2016-11-11
insert callbacks to the back, so that will be invoked in the proper order

morbo84:
We are inserting callbacks to the front of  `onL` and `onceL` lists, but we invoke those callbacks from the `begin()` towards the `end()` of the lists; if we want still to insert callbacks to the front, we should use `rbegin()` and `rend()`, but I think we can safely insert callbacks to the back.

### DIFF
--- a/src/uvw/emitter.hpp
+++ b/src/uvw/emitter.hpp
@@ -46,11 +46,11 @@ class Emitter {
         }
 
         Connection once(Listener f) {
-            return { onceL, onceL.insert(onceL.cbegin(), std::move(f)) };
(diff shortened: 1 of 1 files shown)


skypjack:
This way, it will execute also the callbacks added to the list during the execution of the ones that are already registered. Actually, we already have a similar problem if we consider those callback generated from the execution of the `once` ones.

skypjack:
Updated commit. Is it fine for you as it is?

morbo84:
Well, the initial issue is still there... But I'll found a way to work it out in my client code. Thanks.

PR #72, opened 2016-11-25
Changes as per comments

tusharpm:
The changes are based on the discussion on https://github.com/skypjack/uvw/commit/af5dc89dec39f4f60323808388a94d946fe826d7.

Also, is something wrong with the AppVeyor integration?

### DIFF
--- a/src/uvw/tcp.hpp
+++ b/src/uvw/tcp.hpp
@@ -50,9 +50,7 @@ public:
     using IPv4 = uvw::IPv4;
     using IPv6 = uvw::IPv6;
 
-    explicit TcpHandle(ConstructorAccess ca, std::shared_ptr<Loop> ref)
-        : StreamHandle{std::move(ca), std::move(ref)}, tag{DEFAULT}, flags{}
-    { }
+    using StreamHandle::StreamHandle;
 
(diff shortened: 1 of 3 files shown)


src/uvw/tcp.hpp
```diff
@@ -220,7 +218,7 @@ class TcpHandle final: public StreamHandle<TcpHandle, uv_tcp_t> {
     }
 
 private:
-    enum { DEFAULT, FLAGS } tag;
+    enum { DEFAULT, FLAGS } tag{DEFAULT};
     unsigned int flags;
```
skypjack: I would update also this as `unsigned int flags{};` to be explicit about the fact that all the data members are initialized somehow.

src/uvw/udp.hpp
```diff
@@ -371,7 +369,7 @@ class UDPHandle final: public Handle<UDPHandle, uv_udp_t> {
     }
 
 private:
-    enum { DEFAULT, FLAGS } tag;
+    enum { DEFAULT, FLAGS } tag{DEFAULT};
     unsigned int flags;
 };
```
skypjack: Same as above.

PR #73, opened 2016-11-30
made DataEvent::data unique_ptr non-const

morbo84:
There is no reason to force client code to get read data as an `std::unique_ptr<const char[]>`, and I would change the type of `DataEvent::data` to `std::unique_ptr<char[]>` for two reasons:

- I think it's up to the user whether he wants those data to be modifiable or not
- this way, I could use data received in a `DataEvent` as an argument for `StreamHandle::write()`; on the other hand, if `DataEvent::data` is a _pointer to const_, if I want to write those data through a `StreamHandle` I'm forced to perform an avoidable copy

As a side note: it would be great if the `StreamHandle::write()` function would accept _pointer to const_ as a data argument, but I've not been able to find any guarantee in _libuv_ documentation that those buffer will not be modified by the write (as it should be), so I'm not sure we can `const_cast` our const buffers before they are passed to the `uv_write()` function.


skypjack:
That's clever indeed. Good catch.

PR #77, opened 2017-01-15
Fix GCC warnings -Wshadow

raoulh:
GCC is spawning a lot of warnings about shadowed parameters. This PR fixes all of those.

Renamed shadowed members errors like:
`warning: declaration of ‘path’ shadows a member of ‘uvw::FsEvent<(uvw::details::UVFsType)10>’ [-Wshadow]`

### DIFF
--- a/src/uvw/dns.hpp
+++ b/src/uvw/dns.hpp
@@ -22,8 +22,8 @@ namespace uvw {
 struct AddrInfoEvent: Event<AddrInfoEvent> {
     using Deleter = void(*)(addrinfo *);
 
-    AddrInfoEvent(std::unique_ptr<addrinfo, Deleter> _data)
-        : data{std::move(_data)}
+    AddrInfoEvent(std::unique_ptr<addrinfo, Deleter> addr)
+        : data{std::move(addr)}
     {}
 
     /**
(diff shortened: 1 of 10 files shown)


skypjack:
Thank you for the PR.

Can I ask you for what you are using `uvw`? I'm just curious and glad that someone uses it.

Back to the PR now.
As far as I know it's standard and legal C++, so I'm tempted to reject it. Anyway I see that warnings are annoying for those that use `-Wshadow` and I can accept it.
That said, names don't follow the convention of the project. Do you mind if I change them in the PR directly?

raoulh:
Hi,

Actually I'm porting the server side of the calaos project to use libuv. And because calaos is written in C++, I prefer using your wrapper lib instead of dealing with the C api directly ;)
https://calaos.fr/en
https://github.com/calaos/calaos_base

Our code base is actually built with -Wshadow and uvw triggered lots of warning so here is the PR :)
Sorry for the names that don't follow the convention of the project. You can freely change it thanks!

skypjack:
Wow, awesome.
Let me know your thoughts about `uvw` and feel free to contact me if you have any suggestions for new features and improvements. It goes without saying that PR for bug fixing are welcome!!
If you find it useful, do not forget to star it. It will help to promote the project and attract new developers. :-)

I'm merging the PR during the week.
I'm a bit busy, so expect it to be updated and merged during the second half.
I hope it's not a problem for your team.

Thank you once more.

raoulh:
Sure, thanks for the quick reply :)

PR #78, opened 2017-01-15
Fix build failure when using UDPHandle::recv()

raoulh:
Missing const fails to compile the code with a build error.


skypjack:
Lack of tests. Damn it. Thank you very much.

PR #79, opened 2017-01-15
Fix -Wc++11-narrowing error/warning

raoulh:
Fix a warning with GCC/Linux and a build error with Clang/macOS.

```
udp.hpp:111:82: error: non-constant-expression cannot be narrowed from type 'ssize_t' (aka 'long') to 'std::size_t' (aka 'unsigned long') in initializer list [-Wc++11-narrowing]
        note: insert an explicit cast to silence this issue

udp.hpp:111:89: error: non-constant-expression cannot be narrowed from type 'unsigned int' to 'bool' in initializer list [-Wc++11-narrowing]
        note: insert an explicit cast to silence this issue

udp.hpp:116:82: error: non-constant-expression cannot be narrowed from type 'ssize_t' (aka 'long') to 'std::size_t' (aka 'unsigned long') in initializer list [-Wc++11-narrowing]
        note: insert an explicit cast to silence this issue
```

### DIFF
--- a/src/uvw/udp.hpp
+++ b/src/uvw/udp.hpp
@@ -108,12 +108,12 @@ class UDPHandle final: public Handle<UDPHandle, uv_udp_t> {
 
         if(nread > 0) {
             // data available (can be truncated)
-            udp.publish(UDPDataEvent{details::address<I>(aptr), std::move(data), static_cast<size_t>(nread), static_cast<bool>(flags & UV_UDP_PARTIAL)});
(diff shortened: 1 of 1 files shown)


skypjack:
Thank you for the PR. I think it's definitely time to write tests for UDP handle.
Please, use `std::size_t`. We used the `std::` qualifier all over the code.

raoulh:
Sure, done.

PR #80, opened 2017-01-18
uv_spawn: Do not set cwd to an empty string

raoulh:
If po.cwd is set to an empty string the process failed to start with an error.


skypjack:
I see. Probably libuv treats (correctly) the empty string differently from the null pointer and doesn't accept it. Thank you.

PR #88, opened 2017-07-05
Fixed Watch/Event flags.

filonik:
The Watch/Event flags were mixed up, I have corrected them.

### DIFF
--- a/src/uvw/fs_event.hpp
+++ b/src/uvw/fs_event.hpp
@@ -132,10 +132,10 @@ public:
      * * `FsEventHandle::Event::RECURSIVE`
      *
      * @param path The file or directory to be monitored.
-     * @param watch Additional flag to control the behavior.
+     * @param flag Additional flag to control the behavior.
      */
-    void start(std::string path, Event watch) {
-        start(std::move(path), Flags<Event>{watch});
+    void start(std::string path, Event flag) {
+        start(std::move(path), Flags<Event>{flag});
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/uvw/fs_event.hpp
```diff
@@ -79,7 +79,7 @@ class FsEventHandle final: public Handle<FsEventHandle, uv_fs_event_t> {
     static void startCallback(uv_fs_event_t *handle, const char *filename, int events, int status) {
         FsEventHandle &fsEvent = *(static_cast<FsEventHandle*>(handle->data));
         if(status) { fsEvent.publish(ErrorEvent{status}); }
-        else { fsEvent.publish(FsEventEvent{filename, static_cast<std::underlying_type_t<Event>>(events)}); }
+        else { fsEvent.publish(FsEventEvent{filename, static_cast<std::underlying_type_t<Watch>>(events)}); }
```
skypjack: I suspect it should be `details::UVFsEvent` (for which `Watch` is an alias). `Watch` is defined below, so it shouldn't be a valid name here. It was a bug in a bug.

src/uvw/fs_event.hpp
```diff
@@ -134,8 +134,8 @@ class FsEventHandle final: public Handle<FsEventHandle, uv_fs_event_t> {
      * @param path The file or directory to be monitored.
      * @param watch Additional flag to control the behavior.
      */
-    void start(std::string path, Watch watch) {
-        start(std::move(path), Flags<Watch>{watch});
+    void start(std::string path, Event watch) {
```
skypjack: I would change also the name of the variable. `flag` seems more appropriate, for we are going to pack it in a `Flags` class and invoke the other overload around.

filonik: You are very welcome, uvw is a nice little wrapper. I'd be happy to add my project as soon as I publish the code (so far, it is still in early experimental stages). Keep up the great work!

src/uvw/fs_event.hpp
filonik: Yes, strangely enough, it appears to work (in VS2015). Probably better off changing it though...

PR #93, opened 2017-07-23
CMake and Travis config updates

tusharpm:
* Limit CI jobs to modern versions of each compiler: g++-6 and clang-4.0 on linux. Xcode8.3 for macOS.
* CMake cleanup:
  - Drop minimum version required to 3.2 (available by default on Travis trusty containers). This speeds up the Travis CI builds.
  - Remove implicitly defined variables (like `PROJECT_NAME`, `PROJECT_VERSION`, etc.).
  - Inline some variables which are used only once.
  - Use `endif` consistently -> no expression.
  - Use [`uvw_SOURCE_DIR`](https://cmake.org/cmake/help/v3.2/variable/PROJECT-NAME_SOURCE_DIR.html) instead of `CMAKE_SOURCE_DIR` to allow embedding `uvw` in other CMake projects without patching.
  - Deduplication of boilerplate for all tests -> define and use macros `ADD_UVW_TEST(...)` and `ADD_UVW_DIR_TEST(...)` instead.
---
This is a preparatory step for making `uvw` compatible for embedding and/or packaging using CMake/CPack.
* Next step would be to define `uvw` as an [Interface library](https://cmake.org/cmake/help/v3.2/command/add_library.html#interface-libraries), linking with the underlying `libuv`.
* Afterwards, the `install/export` command can be used to package the required files.


CMakeLists.txt
```diff
@@ -122,11 +90,13 @@ find_library(libuv_shared NAMES uv libuv PATHS ${LIBUV_DEPS_DIR}/src PATH_SUFFIX
...
 
-if(${DOXYGEN_FOUND})
-    add_subdirectory(${PROJECT_DOCS_DIR})
-endif(${DOXYGEN_FOUND})
+if(DOXYGEN_FOUND)
+    add_subdirectory(docs)
+endif()
 
-add_subdirectory(${PROJECT_TEST_DIR})
+if(BUILD_TESTING)
```
skypjack: I suspect `BUILD_TESTING` isn't available with `enable_testing`. Anyway this isn't a good reason to switch to `include(CTest)`, for that variable isn't the sole side effect of such a change. I'm sorry.

test/CMakeLists.txt
```diff
@@ -2,290 +2,74 @@
 # Tests configuration
 #
 
-set(UVW_SRC_DIR ${CMAKE_SOURCE_DIR}/${PROJECT_SRC_DIR})
+set(UVW_SRC_DIR ${PROJECT_SRC_DIR})
+
+if(WIN32)
+    set(WINSOCK2 ws2_32)
+elseif(NOT APPLE)
```
skypjack: Shouldn't you look for `rt` only on Windows? This way it searches it also on linux.

test/CMakeLists.txt
```diff
@@ -2,290 +2,74 @@
...
+    find_library(LIBRT rt)
+endif()
 
 set(
     COMMON_LINK_LIBS
-    gtest
```
skypjack: Just curiosity: why did you remove `gtest`? Is it automatically pulled in by `gtest_main`? I didn't know it works like this.

test/CMakeLists.txt
```diff
@@ -2,290 +2,74 @@
...
+ADD_UVW_TEST(thread uvw/thread.cpp)
+ADD_UVW_TEST(timer uvw/timer.cpp)
+ADD_UVW_TEST(tty uvw/tty.cpp)
+ADD_UVW_TEST(udp uvw/udp.cpp)
+ADD_UVW_TEST(util uvw/util.cpp)
+ADD_UVW_TEST(work uvw/work.cpp)
+
+# Test lib
+
+target_link_libraries(lib PRIVATE ${CMAKE_DL_LIBS})
```
skypjack: I'd rather add a third macro or a dedicated function. Otherwise stuff is spread around the file and it's hard to understand with a glimpse what's going on under the hood.

CMakeLists.txt
```diff
@@ -1,91 +1,60 @@
...
-)
-
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_RUNTIME_OUTPUT_DIRECTORY})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY bin)
 
 #
 # Enable test support using ctest
 #
 
-enable_testing()
```
tusharpm: I'll remove `include(CTest)`.
I'll introduce `BUILD_TESTING` as an option because I'll need that for disabling test builds from the caller.

CMakeLists.txt
tusharpm: Done. Introducing `BUILD_TESTING` without `include(CTest)`.
Moved `enable_testing()` inside the `if` block.

test/CMakeLists.txt
tusharpm: Umm, [`rt` is a posix thing](https://stackoverflow.com/questions/2418157/ubuntu-linux-c-error-undefined-reference-to-clock-gettime-and-clock-settim). It isn't present or required on Windows or macOS.

test/CMakeLists.txt
tusharpm: [It's linked in `gtest_main`](https://github.com/google/googletest/blob/release-1.8.0/googletest/CMakeLists.txt#L92). The [example in googletest's README.md](https://github.com/google/googletest/blame/master/googletest/README.md#L175) uses it this way.

test/CMakeLists.txt
tusharpm: I'm not sure if I understood this one correctly. I've added a third macro `ADD_UVW_LIB_TEST` exclusively for the target `lib`. PTAL.

test/CMakeLists.txt
skypjack: You are right. I'm getting out of a fever that lasted a week. I should just avoid this kind of comments. :-D

skypjack:
Out of curiosity, for what are you using `uvw`?
If you are interested, at the end of the readme file there is a section where you can add your own project or company that uses the project.

test/CMakeLists.txt
skypjack: Good to know. I don't know why, but I was sure both were required. Thank you.

tusharpm:
I'm not using it in production code yet. I've tried integrating it with a couple of pet-projects. Neither of these is published online, hence no links. Sorry.

I understand the header-only part, but my concern is with the use of `uvw` in another (CMake based) project. `libuv` has clearly not adopted CMake, and continues to use gyp. The other project ([libuv_cmake](https://github.com/skypjack/libuv_cmake)) doesn't build with Visual Studio 2017. I'm essentially out of cross-platform options to use.

About the interface library, please visit [the cmake documentation](https://cmake.org/cmake/help/v3.2/manual/cmake-buildsystem.7.html#interface-libraries) to understand more. It **isn't** a library (requiring linking, etc) in the traditional sense. It's a formal definition of an interface and its dependencies, which is a good thing to present in CMake-based solutions (applicable to header-only libraries). A couple of examples of interface libraries (header-only solutions):
* [nlohmann/json](https://github.com/nlohmann/json/blob/develop/CMakeLists.txt#L220)
* [jarro2783/cxxopts](https://github.com/jarro2783/cxxopts/blob/master/CMakeLists.txt#L39)

For the linking to `libuv` part: `cxxopts`, the second example above, has an optional link-time dependency on `icu`. Something similar can be done here to abstract the `libuv` dependency for `uvw`'s users, which in our case is non-optional.

For a larger milestone, `uvw` can provide a cross-platform way to build and link `libuv` along with the CMake buildsystem until they add it internally.

skypjack:
Why `libuv_cmake` doesn't work with VS2017? What error are you getting?  
`uvw` already does its best to provide a cmake based cross-platform way to build and link `libuv` with its tests, and that's exactly what I put in 'libuv_cmake`.

tusharpm:
Well, for VS2017: the reason is [this](https://github.com/libuv/libuv/blob/v1.13.1/vcbuild.bat#L46) and my build goes [here](https://github.com/libuv/libuv/blob/v1.13.1/vcbuild.bat#L72) instead.
Also, I can't build `libuv` using gyp locally anymore because I removed my Python27 installation in favor of Python36. And the way it's built in `libuv_cmake` makes it dependent on both gyp (Python2x) and CMake - making it a non-viable solution. Instead of that, I'd prefer to have an option to disable building `libuv` from `uvw` and using one built/downloaded separately, possibly using a pure CMake build system [like this one](https://github.com/jen20/libuv-cmake/blob/v1.9.0/CMakeLists.txt), or a package manager eg. vcpkg for Windows.

skypjack:
I see. Pure cmake build systems for `libuv` can be annoying to maintain and I avoided them. `libuv` devs explicitly said they won't support any solution and a maintainer is required in any case. I decided to use their build system directly then.

Anyway, py2.7 and py3.6 can live together in a system. I'm with Debian Sid and I've both of them. At the end of the day you are going to solve a problem you imposed to yourself. :-)

I'm at your disposal for help if needed in any case.

tusharpm:
At least the issue "Add CMake support" is open (for now). Hopefully someday someone there will get rid of gyp and decide to move on.
I agree the two versions of Python can coexist (as they do on AppVeyor's machines). But it was creating problems for another project (coincidentally, the project where I eventually want to use `uvw`) which led me to remove the older one.

In any case, I thought the file `libuv.in` should be aware of the new command-line argument to `vcbuild.bat` and use it when required.

skypjack:
>I'd prefer to have an option to disable building libuv from uvw and using one built/downloaded separately

We could move the stuff for `gtest` and `libuv` within the `if`, so that it depends on `BUILD_TESTING` and you can easily exclude them. Actually all for what they are required are the tests.

skypjack:
>In any case, I thought the file libuv.in should be aware of the new command-line argument to vcbuild.bat and use it when required.

Feel free to propose patches, you're welcome!! One of the major contributors of `uvw`... ;-)

PR #94, opened 2017-07-24
Common create method

tusharpm:
The advantage of having a common `create` method is that it consolidates the dynamic allocation of "resources" in `uvw`. This doesn't exhaust all instances, but reduces the usage of `new` in existing code.
An observable benefit of the change is the reduction of duplicate arguments to the `create` method and the class constructors.

**Change description:**
Consolidate the `<Type>::create` methods from various classes - `SharedLib`, `Thread` and others in a single base class. Also aggregate other similar functionality like getting the owning `Loop` object reference, getting the owned libuv structure pointer.
Since `class Resource` inherits `Emitter` and `std::enable_shared_from_this`, it isn't a feasible option. Therefore, `ResourceBase` (the name can be decided upon) is extracted from and inherited into `Resource`. This essentially means no API change in the final classes of `uvw`.

### DIFF
--- a/src/uvw/lib.hpp
+++ b/src/uvw/lib.hpp
@@ -1,12 +1,13 @@
 #pragma once
 
 
-#include <utility>
 #include <memory>
 #include <string>
 #include <type_traits>
+#include <utility>
 #include <uv.h>
 #include "loop.hpp"
+#include "resource_base.hpp"
 
 
 namespace uvw {
@@ -18,13 +19,7 @@ namespace uvw {
  * `uvw` provides cross platform utilities for loading shared libraries and
  * retrieving symbols from them, by means of the API offered by `libuv`.
  */
-class SharedLib final {
-    explicit SharedLib(std::shared_ptr<Loop> ref, std::string filename) noexcept
-        : pLoop{std::move(ref)}, lib{}
-    {
-        opened = (0 == uv_dlopen(filename.data(), &lib));
-    }
-
+class SharedLib final: public ResourceBase<SharedLib, uv_lib_t> {
 public:
     /**
      * @brief Creates a new shared library object.
@@ -32,19 +27,15 @@ public:
      * @param filename The filename of the library in UTF8.
      * @return A pointer to the newly created handle.
      */
-    static std::shared_ptr<SharedLib> create(std::shared_ptr<Loop> loop, std::string filename) noexcept {
-        return std::shared_ptr<SharedLib>{new SharedLib{std::move(loop), std::move(filename)}};
+    explicit SharedLib(ConstructorAccess ca, std::shared_ptr<Loop> ref, std::string filename) noexcept
+        : ResourceBase{std::move(ca), std::move(ref)}
+    {
+        opened = (0 == uv_dlopen(filename.data(), get()));
     }
-
-    SharedLib(const SharedLib &) = delete;
-    SharedLib(SharedLib &&) = delete;
 
     ~SharedLib() noexcept {
-        uv_dlclose(&lib);
+        uv_dlclose(get());
     }
-
-    SharedLib & operator=(const SharedLib &) = delete;
-    SharedLib & operator=(SharedLib &&) = delete;
 
     /**
      * @brief Checks if the library has been correctly opened.
@@ -65,7 +56,7 @@ public:
     F * sym(std::string name) {
         static_assert(std::is_function<F>::value, "!");
         F *func;
-        auto err = uv_dlsym(&lib, name.data(), reinterpret_cast<void**>(&func));
+        auto err = uv_dlsym(get(), name.data(), reinterpret_cast<void**>(&func));
         if(err) { func = nullptr; }
         return func;
     }
@@ -75,18 +66,10 @@ public:
      * @return The last error message, if any.
      */
     const char * error() const noexcept {
-        return uv_dlerror(&lib);
+        return uv_dlerror(get());
     }
 
-    /**
-     * @brief Gets the loop from which the object was originated.
-     * @return A reference to a loop instance.
-     */
-    Loop & loop() const noexcept { return *pLoop; }
-
 private:
-    std::shared_ptr<Loop> pLoop;
-    uv_lib_t lib;
     bool opened;
 };
 
--- a/src/uvw/lib.hpp
+++ b/src/uvw/lib.hpp
@@ -7,7 +7,7 @@
 #include <utility>
 #include <uv.h>
 #include "loop.hpp"
-#include "resource_base.hpp"
+#include "underlying_type.hpp"
 
 
 namespace uvw {
@@ -19,7 +19,7 @@ namespace uvw {
  * `uvw` provides cross platform utilities for loading shared libraries and
  * retrieving symbols from them, by means of the API offered by `libuv`.
  */
-class SharedLib final: public ResourceBase<SharedLib, uv_lib_t> {
+class SharedLib final: public UnderlyingType<SharedLib, uv_lib_t> {
 public:
     /**
      * @brief Creates a new shared library object.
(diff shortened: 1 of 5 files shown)


skypjack:
I've no time to look over this today. I'm sorry. I'm a bit busy at work.  
I'll review the changes during the week and I'll be back with comments if needed.

skypjack:
>Since class Resource inherits Emitter and std::enable_shared_from_this, it isn't a feasible option.

I don't get the reasons for that.  
By defining a couple of sfinae'd private constructors to which to forward everything from a more generic constructor (not all `*_t` structs of `libuv` have a `data` member) we can safely use `Resource` as a common base class.  
`Emitter` will give the chance to use those classes to emit events (namely as observables).  
`std::shared_from_this` can help offering the same `leak`/`reset` functionalities used by handles and requests.

I agree on the reasons for which you opened the PR. Anyway I'm not sure splitting up `Resource` in two classes is the best way to solve it.

If you agree, I'll have a try to propose an alternative approach before to accept or reject the PR. Give me a few days to arrange it.

tusharpm:
The classes modified here don't "emit" events as their functionality is always synchronous and doesn't involve the event loop, hence no need of the observer pattern. This also implies that they don't require `leak`/`reset` for modifying ownership like handles or requests. That's why I didn't inherit those.
I'd prefer a SFINAE approach if it can achieve the same objective.
I am not opposed to trying a different approach; I'd just like to get this clarified before you make that change.

skypjack:
I mean something along this line: [resource.hpp](https://github.com/skypjack/uvw/blob/common_resource/src/uvw/resource.hpp) and [lib.hpp](https://github.com/skypjack/uvw/blob/common_resource/src/uvw/lib.hpp).
Do not forget that member methods of a template class aren't instantiated unless you use it, so it's safe to ignore `leak`, release` and the others if we don't use them. On the other side, `SharedLib` is now an emitter as all the other classes in `uvw` are, no matters if the users will use or not this feature.

Let me know if you need more details. I'll look forward for your thoughts.

tusharpm:
Quick as always :smile:.
I was actually trying to emphasise the point of not having something we know would not work - like the `Emitter` members inherited in `SharedLib` and others here. These functions (`on` or `once`) even if called, would not have any effect i.e. `SharedLib` or `Mutex` are not going to invoke event callbacks, which I think is more clearly stated by not having them available.

This being a possible design decision: if you feel having these interfaces is useful for these classes, please proceed with your changes.

My comment on your change - maybe you don't need to replicate the whole constructor and the member initialiser-list. Keep them in the same constructor; just move the one line from the body of the constructor to a different template function.

skypjack:
Yeah, `Emitter` isn't used directly by those classes. Anyway, if I'm using a library where almost all classes are designed around a P&S model, I usually expect all of them to be event-aware. Whenever I need to emit events through an instance for some reasons, I can do that.  
As an example, imagine I create a kind of library manager class that is in charge to load and unload libraries and use the `SharedLib` class as a sink on which to attach listeners to notify all those interested in this stuff. I can create my own sink, of course, but... well, handles and requests are emitters, why did you decide not to make emitters also the other ones? Only becuase `libuv` treats them in a different way? Uhm...

Hope I succeeded giving you my point of view. It's more a matter of taste in the design of the API, not a problem of functionalities.

tusharpm:
> almost all classes are designed around a P&S model

This is going to sound nitpicking, but the keyword I'm pointing to is _almost_. Not all classes are using the publish-subscribe mechanism. Especially not the ones being modified here. Furthermore, since `Emitter::publish` is `protected` and every class meant for the `uvw` users is `final`, it doesn't look like there's going to be external events published from this interface. This essentially also rules out using a `SharedLib` powered library manager with notifications.
> Whenever I need to emit events through an instance for some reasons, I can do that.

That argument brings in a future component, which I'm not sure about. So far, events have always been emitted in a callback from the `libuv`'s event loop (the only exception is `ErrorEvent`, please see below for its explanation). Unless the plan is to "fabricate" events without updates from `libuv`, I don't think the situation is going to change.
> handles and requests are emitters, why did you decide not to make emitters also the other ones?

Actually, I didn't. The others are not `Emitter`s in existing code. I just thought it was better to maintain the same behavior.

About `ErrorEvent`, it's a dual-purpose event which is emitted synchronously for some errors from `libuv`. That is one behavior that could possibly not be abstracted for handles/requests, but, unlike those errors in the classes being modified here are clearly `return`ed. Unless the plan is to modify the external interface to pass errors from these classes also as `ErrorEvent`s (which would also make these more verbose to use), it can be ruled out of the argument.

---
I understand your point of view. I'm sure you would've understood mine. I believe you're best-equipped to take a call on which way is best for `uvw`.

skypjack:
>Furthermore, since `Emitter::publish` is protected [...]

Good point. I didn't remember it. Actually it makes inheriting from `Emitter` pointless.

>I understand your point of view. I'm sure you would've understood mine. I believe you're best-equipped to take a call on which way is best for uvw.

Sorry, I didn't want to bother you. I appreciate your help and each and every time I have the chance to discuss different points of view I do that. They are so precious, the good ones as well as the wrong ones.

---

As a side note, before to look for a way to better integrate threads stuff, it's worth asking if it shouldn't be eliminated. I mean, C++ has good support for threads out-of-the-box with `<thread>`. Does it make sense at all to keep pulling in thread stuff from `libuv`? I'd rather encourage people to use the native way instead.  
What do you think about it?

tusharpm:
I apologise if my response came across as confrontational; that is not my intent. I try to put my opinions objectively, which at times makes them sound blunt. My statement was meant as a gesture of appreciation for your opinion, and your veto right as creator of `uvw`.

---
I agree the standard thread library and synchronisation primitives could be recommended in place of the ones we get from `libuv` wherever possible. Things to verify before that's done:
* Check if it affects `libuv`'s functionality like thread-pool, or consequentially async/work handles.
* Check if all interfaces are available. For instance, IIRC `Once`, `RWLock`, `Semaphore` or `Barrier` don't have alternatives in the standard library. The last one appears in concurrency TS. For the others being eliminated, there should be a suitable replacement available.

skypjack:
Quite true, you are right. Well, let's integrate the patch and see how it works.  
The only thing remained to do is to find a better name than `ResourceBase`!! :-)
I really appreciate your help and feedbacks, anyway `ResourceBase` is a terrible choice... :-D

tusharpm:
I agree, `ResourceBase` is possibly the laziest choice of class names. 😅
Now that I think about it, I can't come up with an informative name. My suggestions:
- `Creatable` = this describes the interface it's meant to be.
Cons:
  - Seems like a concept name.
  - Doesn't exclude `Loop` which is also "creatable", but not derived from this type.
- `InternalResource`
Cons:
  - I'm not sure it's a step up from `ResourceBase` or `BaseResource` or the like.
  - This might create more confusion.

If you have any suggestions, please let me know.

skypjack:
`Concept`? `Basic`? `UnderlyingType`? Maybe the latter is explicit on the fact that that class is nothing more than a wrapper around a resource defined by `libuv`. Actually, it does nothing but to wrap an `uv_*_t` type, right?

skypjack:
@tusharpm Can you adapt the code to the _new_ name? If you can't, that's not a problem. Absolutely. I can merge the PR and proceed with the changes. Let me know if you are already working on it. Thanks.

tusharpm:
I pushed a commit renaming `ResourceBase` to `UnderlyingType`. If you could tell me what else needs changing, I can try to incorporate that as well.

skypjack:
Wanna know the funny part? I've just seen the commit you made 3 days ago. :-D

skypjack:
Thank you very much. Sorry if I'm late with the merge, but I'm a bit busy at work.

skypjack:
@tusharpm `friend` declaration and `get` member functions that accepted a `Resource` as an argument have been accidentally removed by this PR and the `Process` class is broken now.  
Unfortunately it lacks tests, so we didn't notice that.  
We should put back them in the `UnderlyingType` class. I can do that in the evening.  
Is there a reason for which you removed them?

tusharpm:
Oh, what a blunder. My mistake. Totally missed that.
Removing the `static` `get` members was intentional, as the non-`static` `get` members are less "imperative", and appear clearer in specifying intent.
However, using those methods requires the `friend` specifier. Removing that was accidental. After moving the code around (part of which I clearly missed) I verified the existing tests only, which got built and passed.
My apologies.

skypjack:
Not a problem. I'm going to fix it, no need for a PR.  
It happens. Unfortunately the test suite doesn't fully cover each class yet, my fault.

PR #97, opened 2017-08-14
Fix compilation on win32

slyshykO:


CMakeLists.txt
```diff
@@ -34,9 +34,9 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 if(NOT WIN32)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -g -DDEBUG")
```
skypjack: We already have `appveyor` up and running to test compilation on Windows at each commit.  
There is no need to move the lines above within the `if` and exclude them when running on Windows.  
What's exactly the problem you are trying to solve?  
This change just breaks the build system, it doesn't look a good one honestly.

src/uvw/stream.hpp
```diff
@@ -155,8 +155,8 @@ class StreamHandle: public Handle<T, U> {
 
 public:
 #ifdef _WIN32
-    StreamHandle(ConstructorAccess ca, std::shared_ptr<Loop> ref)
-        : Handle{ca, std::move(ref)}
+    StreamHandle(typename Handle<T, U>::ConstructorAccess ca, std::shared_ptr<Loop> ref)
+        : Handle<T, U>{std::move(ca), std::move(ref)}
```
skypjack: What problem should this solves exactly? What error are you having?  
Set aside the fact that `ConstructorAccess` is a tag that doesn't require to be moved, it's already a custom part aimed to get compilation working on Windows and it just works on `appveyor` as expected.  
As far as I can see, this change isn't required.

CMakeLists.txt
slyshykO: I have a question, which compiler you use in appveyor?
I thought it is msvc, according to the content of libuv.in . But -O0, -g and -Os it's not correct options for msvc.
Maybe it should move to proper _if_ , for example:
`if(NOT MSVC`  ?

src/uvw/stream.hpp
slyshykO: I agree that `std::move(ca)` is not necessary. But other changes need for success compilation under mingw

CMakeLists.txt
skypjack: MSVC just ignores them if not used and compiles tests correctly, so I wouldn't add boilerplate to solve a problem that doesn't exist.  
At the end of the day, maintainability is a good thing. ;-). 
Moreover, `WIN32` doesn't mean MSVC, so this change is wrong in any case. Are you experiencing an actual problem or is the change a premature optimization? In the first case, what error are you encountering?

src/uvw/stream.hpp
skypjack: I suspect that the problem is [GCC version provided with MinGW](http://www.mingw.org/wiki/GCCStatus). `uvw` requires a fully compliant C++14 compiler, GCC4.4 isn't one of them. I suspect you'll encounter more problems in future.  
In order to accept this change we should at least add GCC4.4 to Travis (see `.travis.yml`) and verify that all the tests compile and run correctly. Did you try to compile and run them under MinGW?

slyshykO: I use MinGW-W64 GCC-7.1.0 for build my project.
https://sourceforge.net/projects/mingw-w64/files/

skypjack: Interesting. What error are you experiencing without those changes? What's the error message? GCC7 should compile it just fine.

slyshykO: `
../../swdt/./3rdparty/uvw/uvw/stream.hpp:158:36: error: expected ')' before 'ca'

     StreamHandle(ConstructorAccess ca, std::shared_ptr<Loop> ref)

                                    ^~

../../swdt/./3rdparty/uvw/uvw/stream.hpp:160:5: error: expected unqualified-id before '{' token

     {}

     ^
`
gcc can't resolve **ConstructorAccess** and **Handle** types. 
Look, in non WIN32 version of code you use **Handle<T, U>::Handle** ,  not simple **Handle**

skypjack:
Got it. Probably the best isn't to modify the win version and to struggle with MSVC idiosyncrasies. Instead, the `if` should probably discriminate on MSVC instead of WIN32.  
Have you the chance to test it? Unfortunately I'm out of home for a few days and I've not my laptop with me.

skypjack:
Just a question: how is it that MinGW defines `_WIN32`? It should define `__MINGW32__` instead, am I wrong?

slyshykO:
**_WIN32** defined for all compilers for windows. **__MINGW32__** defined in addition to so.

src/uvw/stream.hpp
```diff
@@ -155,8 +155,8 @@ class StreamHandle: public Handle<T, U> {
 
 public:
 #ifdef _WIN32
```
tusharpm: This can be changed to `#ifdef _MSC_VER` instead. Or if you know which version of MSVC supports the inherited constructor form, `#if defined(_MSC_VER) && _MSC_VER < supported_version`.
The block here was added for not-fully-compliant compilers - primarily MSVC. GCC should be fine with the `#else` block.

skypjack:
Good point with `_MSC_VER`. @slyshykO can you try it? Otherwise I'll patch `uvw` once back from vacation.

slyshykO:
That code compiles if apply `#ifdef _MSC_VER` by mingw-64 gcc. But despite that inherited constructor declared for msvc 15 this one fall to error 

> stream.hpp:162: error: C2385: ambiguous access of '{ctor}'

skypjack:
Uhm. It shouldn't break out this way. Guess what's wrong.  
I plan to look over the issue as soon as I'm back from vacation.  
Hope it's not a problem for you to wait until the end of August.

slyshykO:
Ok, no problem

src/uvw/stream.hpp
skypjack: @tusharpm 
OP says that changing it to `#ifdef _MSC_VER` works fine for MinGW, but it breaks out MSVC.  
I've not a Win env to test it, set aside appveyor. Do you have any chance to look over the change and find what's wrong?  
Let me know. Thank you very much in any case.

tusharpm: For the last commit, I need to request to revert all changes inside these blocks - I'll do that in a separate comment. Essentially nothing changes for MSVC. Otherwise, there's nothing wrong with it. The AppVeyor history for your repo has [one successful build for the same](https://ci.appveyor.com/project/skypjack/uvw/build/1.0.243).

If I understood @slyshykO's response correctly, it would break if we used the `#else` block (specifically line 162) for newer MSVC like in the second suggestion in my previous comment. Requesting @slyshykO to correct me if I got it wrong.

IMO, we can stick to the `#ifdef _MSC_VER` form if it is acceptable.
I checked it on my machine: the existing tests build and run with `#ifdef _MSC_VER`.

src/uvw/stream.hpp
```diff
@@ -154,9 +154,9 @@ class StreamHandle: public Handle<T, U> {
     }
 
 public:
-#ifdef _WIN32
-    StreamHandle(ConstructorAccess ca, std::shared_ptr<Loop> ref)
-        : Handle{ca, std::move(ref)}
+#ifdef _MSC_VER
+    StreamHandle(typename Handle<T, U>::ConstructorAccess ca, std::shared_ptr<Loop> ref)
+        : Handle<T, U>{ca, std::move(ref)}
```
tusharpm: These lines (158-159) should not need to be changed - unless you're getting an explicit error from MSVC. It should be able to deduce these expansions implicitly.

slyshykO: done.

skypjack:
Thank you very much for the PR. Sorry if I'm late, I was on vacation. Next time the review will be faster.

PR #100, opened 2017-09-11
return eagain if nread is 0 per libuv

bmagistro:
Currently if nread is 0 it is treated as an error with code 0.  This is bubbled up in the ErrorEvent as an "unknown system error" (for me).  The docs for libuv indicate that nread can return 0 and it does not indicate an "error" or EOF but that it is equivalent to EAGAIN/EWOULDBLOCK.  Update changes the behavior to pass EAGAIN as the error code if nread is 0 otherwise return the value provided in nread.

### DIFF
--- a/src/uvw/stream.hpp
+++ b/src/uvw/stream.hpp
@@ -10,7 +10,6 @@
 #include "request.hpp"
 #include "handle.hpp"
 #include "loop.hpp"
-#include <cerrno>
 
 
 namespace uvw {
@@ -142,15 +141,13 @@ class StreamHandle: public Handle<T, U> {
         } else if(nread > 0) {
             // data available
             ref.publish(DataEvent{std::move(data), static_cast<std::size_t>(nread)});
+        } else if(nread == 0) {
+            // equivalent to EAGAIN/EWOULDBLOCK
+            // http://docs.libuv.org/en/v1.x/stream.html
+            // simply catch this condition and do not publish an event
         } else {
-            if (nread == 0) {
-                // equivalent to EAGAIN/EWOULDBLOCK
-                // http://docs.libuv.org/en/v1.x/stream.html
-                ref.publish(ErrorEvent(EAGAIN));
-            } else {
-                // transmission error
-                ref.publish(ErrorEvent(nread));
-            }
+            // transmission error
+            ref.publish(ErrorEvent(nread));
         }
     }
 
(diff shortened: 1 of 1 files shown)


skypjack:
Good spot the fact that `nread` set to 0 isn't an error. Anyway, I don't think that it should raise an `ErrorEvent`, because... it's not an error!! Also `libuv` explicitly says to not consider it an error indeed.

Options are:

* Silently ignore it: `uvw` abstracts away the low-level details of the underlying library, probably it should go for it also this time.

* Emit a `DataEvent` with no data. In other terms, propagate to the caller the same _result_ and follow the same pattern.

`ErrorEvent`s should be followed by a `stop` or a `close` (according to the `libuv` documentation), that is not the case for `nread` set to 0.
I cannot accept this PR, I'm sorry.

bmagistro:
All of those sound like things that we thought at different times too.  I was mixed about returning an `ErrorEvent` as well but it fit with the checks we are using now `err.code() == 0 && (errno == EAGAIN || errno == EWOULDBLOCK)` to skip the stop/close.  While not an "error", I'm not sure how else you would provide that notification back so that leaves two options, a DataEvent or masking the occurrence.

I am not wild about returning it as a DataEvent with no data.  From the `write(2)` man page.
> If count is zero and fd refers to a file other than a regular file, the results are not specified.

If trying to say act as a proxy for two connections and you use the `on DataEvent` to trigger a write to the other connection checks MUST be added to check data length before continuing.  This seems like it could be risky for other code already out there.

This leaves it with hiding this detail.  I am not sure anyone would do anything with the information that `EAGAIN` occurred so masking it internally is likely the cleanest option.  It also keeps the DataEvent behavior the same as it is today.

Thoughts?

skypjack:
>I am not sure anyone would do anything with the information that EAGAIN occurred

I agree. Pretty much something to ignore my side.  
Anyway here is another solution that is half way:

* Hides the fact that `nread` is 0 along the _standard path_ (no `DataEvent`, no `ErrorEvent`) so that we won't break existent code

* Emit a dedicated event (not sure what should be the name, probably something like `WouldBlockEvent`) for who knows what to do with it

What about this?

skypjack:
After thinking about it for a while, I'd say that suppress it seems the best choice.
That's such a low level detail I can't see any practical use in this stage.
Moreover, because of how `libuv` works, there is no reason to notify the users of `uvw`.

My two cents. If you agree, I'm merging the PR as-is on `master`.

bmagistro:
I'm good with just suppressing it.  Thanks!

skypjack:
You're welcome. I'm adding you to the AUTHORS file.
For you have not your real name on github, I'm using the sole alias.

PR #101, opened 2017-09-24
remove unnecessary using statement

bmagistro:
While trying to run our code through some static analysis tools, we were getting an error running.  It was eventually traced back to the line below.  Comparing with tty and pipe as well as an attempt to remove and retest our code leads us to believe this line is unnecessary.

Thoughts on removing it?

### DIFF
--- a/src/uvw/tcp.hpp
+++ b/src/uvw/tcp.hpp
@@ -49,8 +49,6 @@ public:
     using IPv4 = uvw::IPv4;
     using IPv6 = uvw::IPv6;
 
-    using StreamHandle::StreamHandle;
-
     explicit TcpHandle(ConstructorAccess ca, std::shared_ptr<Loop> ref, unsigned int f)
         : StreamHandle{ca, std::move(ref)}, tag{FLAGS}, flags{f}
     {}


skypjack:
As you can see, [it isn't unnecessary](https://travis-ci.org/skypjack/uvw/builds/279237464). This PR cannot be accepted as it stands. I'm sorry.

skypjack:
The best you can do to remove the using declaration is:

* Change `enum { DEFAULT, FLAGS } tag{DEFAULT};` in `enum { DEFAULT, FLAGS } tag;`

* Rewrite the constructor as:
    ```
    explicit TcpHandle(ConstructorAccess ca, std::shared_ptr<Loop> ref, unsigned int f = {})
        : StreamHandle{ca, std::move(ref)}, tag{f ? FLAGS : DEFAULT}, flags{f}
    {}
    ```

If I'm right, now you can remove it.

skypjack:
Well, for it made sense, I've just pushed the changes in the comment above. See 3b28f21.

bmagistro:
Ya, I saw the build but wasn't quite sure how to go about fixing things so was going to look at it with a couple other folks today.

I can also confirm this change makes the static analysis tools we were trying to run happy again (but also shows I've got some work to do...).  Thanks for working though this!!

Issue #102, opened 2017-11-22
Improve low-level TCP connection interface


eklitzke:
I have an application that needs to do the following:
 * Resolve a DNS name
 * Try to connect to all the resolved addresses (until one succeeds)

This is currently difficult to do with the API exposed by uvw. The DNS interface is just fine, but I think the TCP connect methods need some work. Specifically, the TCP::connect methods take a string name, and only connect to the first resolved name.

Right now I've just kind of copied the logic from TCP::connect in my own application, which works but is kind of ugly (in my opinion). A cleaner way to do this would be to add a new TCP::connect overloaded method that accepts a pointer/reference to a struct addrinfo.

skypjack:
I'm out of home until Sunday. I'll look into the issue next week probably.  
I'm sorry for the inconvenience.

skypjack:
Let me know if it works for you. It's on branch `wip`.  
If it's fine, I can merge it on `master` and close the issue.

skypjack:
[Available on `master`](https://github.com/skypjack/uvw/blob/master/src/uvw/tcp.hpp#L208).

PR #103, opened 2017-11-22
Add a TCP::connect overload for addrinfo structs

eklitzke:
This is my proposed solution to #102 .


skypjack:
I'm out of home until Sunday. I'll look into the PR next week probably.  
I'm sorry for the inconvenience.

eklitzke:
No worries, take your time.

skypjack:
I wouldn't do that probably. You don't need to pull in another function template actually.  
We can rework it introducing a new `connect` that accepts an already erased addr type. The other `connect` around can just `reinterpret_cast` it before to push the parameter to the new function.

Would it work for you?

eklitzke:
That sounds good to me.

PR #104, opened 2017-11-25
Implement equality checks and hashing for uvw::Addr

eklitzke:
Pretty self-explanatory.

For reasons that are not clear to me, line 172 of util.hpp (which checks that the set has size 1) generates a compiler warning (GCC 7.2.1):

```
In file included from /home/evan/code/uvw/test/uvw/util.cpp:4:0:
/home/evan/code/uvw/deps/googletest/src/googletest/include/gtest/gtest.h: In instantiation of ‘testing::AssertionResult testing::internal::CmpHelperEQ(const char*, const char*, const T1&, const T2&) [with T1 = long unsigned int; T2 = int]’:
/home/evan/code/uvw/deps/googletest/src/googletest/include/gtest/gtest.h:1421:23:   required from ‘static testing::AssertionResult testing::internal::EqHelper<lhs_is_null_literal>::Compare(const char*, const char*, const T1&, const T2&) [with T1 = long unsigned int; T2 = int; bool lhs_is_null_literal = false]’
/home/evan/code/uvw/test/uvw/util.cpp:172:5:   required from here
/home/evan/code/uvw/deps/googletest/src/googletest/include/gtest/gtest.h:1392:11: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (lhs == rhs) {
```

The warning is harmless, but I couldn't figure out how to suppress it.

### DIFF
--- a/src/uvw/util.hpp
+++ b/src/uvw/util.hpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <stdexcept>
 #include <cstddef>
+#include <functional>
 #include <utility>
 #include <string>
 #include <vector>
@@ -299,6 +300,10 @@ struct IPv6 {};
 struct Addr {
     std::string ip; /*!< Either an IPv4 or an IPv6. */
     unsigned int port; /*!< A valid service identifier. */
+
+    bool operator==(const Addr &other) const {
+        return ip == other.ip && port == other.port;
+    }
 };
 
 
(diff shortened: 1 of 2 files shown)


skypjack:
Any feedback? I'm going to close the PR otherwise for I've not received an answer so far and the problem seems to be in the new test, not in the class itself.

skypjack:
I'm closing this for the error has been introduced by a test that isn't upstream.
Please, fix the test and try again. If the problem persists, feel free to update the PR and reopen it.
Thank you.

PR #109, opened 2018-01-30
FEATURE: expose write_queue_size

bmagistro:
There may be a better way to access member variables from the underlying libuv types but I'm not sure how to do it.

### DIFF
--- a/src/uvw/stream.hpp
+++ b/src/uvw/stream.hpp
@@ -447,11 +447,11 @@ public:
     }
 
     /**
-     * @brief Warpper to access uv_stream_t.write_queue_size
+     * @brief Retrieves stream->write_queue_size
      * @return Amount of queued bytes waiting to be sent
      */
-    size_t get_write_queue_size() const noexcept {
-        return this->template get<uv_stream_t>()->write_queue_size;
+    size_t writeQueueSize() const noexcept {
+        return uv_stream_get_write_queue_size(this->template get<uv_stream_t>());
     }
 };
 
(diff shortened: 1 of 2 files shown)


skypjack:
Do you mean `size_t uv_stream_get_write_queue_size(const uv_stream_t* stream)`?

Moreover, the method you defined doesn't follow the naming convention of the whole project.
I can't accept it as it is, I'm sorry.

bmagistro:
I am not sure what you are referring to.  In this case I am referring to `uv_stream_t.write_queue_size` as defined here http://docs.libuv.org/en/v1.x/stream.html.  I can easily rename the function to be `writeQueueSize()`.

skypjack:
`uv_stream_get_write_queue_size` is a new method introduced with libuv 1.19 to get that parameter in a _safe_ way.

bmagistro:
Huh, thanks for pointing that out.  Now I'm not sure I ever scrolled down to the bottom of the page (or I've just been missing it).  I will update to use this method and update name accrodingly.

skypjack:
You're welcome.
I'd name it with a shorter name probably, something like `queue` and that's all. I don't see neither a `read_queue_size` nor anything else on `libuv`, so being that explicit and redundant is probably useless. My two cents.

bmagistro:
So I did a little looking to see what else might be new in the parts I'm using.  In UDP there are `send_queue_count` and `send_queue_size`, so went with the name mentioned above.

skypjack:
Can you add also the other functions from UDP in the PR?

bmagistro:
Functions added, guess I forgot to hit comment the other day.

skypjack:
Yeah, I saw the commit. I have to add a few testsbefore to merge it but I've been busy today and I didn't do that. I'm sorry.
Hopefully tomorrow, otherwise I'll tackle it in during the weekend. ;-)

PR #115, opened 2018-06-24
Small fixes in CMake

janekolszak:
Adding uvw as an external project was problematic because:
- it had to run configure scripts on each build.
- it hasn't installed the headers


skypjack:
Thank you. Upstream. I'm not creating a tag right now, mainly because there is `libuv` v1.21 to port to `uvw` and it's worth creating a single tag once done.

PR #116, opened 2018-07-08
Fix cross-compilation

janekolszak:
Cross compilation for Jetson board was failing with:

CMake Error at /usr/local/share/cmake-3.11/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Threads (missing: Threads_FOUND)
Call Stack (most recent call first):
  /usr/local/share/cmake-3.11/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/share/cmake-3.11/Modules/FindThreads.cmake:205 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:64 (find_package)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,7 @@ option(BUILD_TESTING "Enable testing with ctest." ON)
 
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 
-find_package(Threads REQUIRED)
+include(FindThreads)
 find_package(Doxygen 1.8)
 
 #


CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ option(BUILD_TESTING "Enable testing with ctest." ON)
 
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 
-find_package(Threads REQUIRED)
+include(FindThreads)
```
skypjack: Just out of curiosity (I'm not a cmake expert, so I'd like to get the chance to learn something).  
What's the difference between the two lines (`find_package` and `include` both of threads module)?

janekolszak: find_package just sets some variables and ensures package is present

FindThreads is a module that will respect the THREADS_PREFER_PTHREAD_FLAG variable you added. On my ARM target it seems to be more clever than find_package.

PR #123, opened 2018-08-18
CMake updates

tusharpm:
CMake is a supported build system for libuv since version 1.21.
That can be leveraged to simplify the uvw test build.

Includes an optimisation to reduce one CMake invocation for fetching dependencies.
Note: for local builds the `deps` directory should be cleaned up first (`git clean -ffdx deps`).


skypjack:
It would be great, but something failed. Can you look into the problem?

tusharpm:
It had failed in one of the libuv tests, [time precision of uv_sleep](https://github.com/libuv/libuv/blob/v1.22.0/test/test-hrtime.c#L50).

I disabled running the tests for the dependencies.

skypjack:
Oh, sorry, I'm at the beach with my son, I didn't see what failed. Forgive me.

skypjack:
Thank as usual.

PR #131, opened 2018-10-05
Follow ctest and BUILD_TESTING best practices

### LINKED ISSUE #130: Please don't default BUILD_TESTING to ON - it messes with other projects
`BUILD_TESTING` is considered a 'special' option for the entire CMake environment, and is thus used across all projects that use CMake.

Explicitly defining it using `option()` is unnecessary (using it in an `if()` when it hasn't explicitly been set just results in a falsey conditional value) -- and further, specifying it as `ON` causes drastically higher compilation times for those that do not want them.

Qix-:
Closes #130.

Thanks for the swift response :) Always nice when that happens.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -26,7 +26,7 @@ macro(ADD_UVW_TEST TEST_NAME TEST_SOURCE)
     add_executable(${TEST_NAME} $<TARGET_OBJECTS:odr> ${TEST_SOURCE})
     target_include_directories(${TEST_NAME} PRIVATE ${UVW_SRC_DIR})
     target_link_libraries(${TEST_NAME} PRIVATE ${COMMON_LINK_LIBS})
-    add_test(NAME ${TEST_NAME} COMMAND ${TEST_NAME})
+    add_test(NAME uvw_${TEST_NAME} COMMAND ${TEST_NAME})
 endmacro()
 
 macro(ADD_UVW_DIR_TEST TEST_NAME TEST_SOURCE)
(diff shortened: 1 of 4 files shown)


skypjack:
Is it fine for you if I change `On` to `ON`? I can directly edit your commit.

Qix-:
Sure, no problem. I don't think it matters either way.

skypjack:
>Sure, no problem. I don't think it matters either way (as in, CMake will accept either).

Yeah, it's me that prefer to be consistent through the whole project. Thanks.

test/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ macro(ADD_UVW_TEST TEST_NAME TEST_SOURCE)
     add_executable(${TEST_NAME} $<TARGET_OBJECTS:odr> ${TEST_SOURCE})
     target_include_directories(${TEST_NAME} PRIVATE ${UVW_SRC_DIR})
     target_link_libraries(${TEST_NAME} PRIVATE ${COMMON_LINK_LIBS})
-    add_test(NAME uvw-${TEST_NAME} COMMAND ${TEST_NAME})
```
Qix-: Please keep the prefix. It helps when running tests from several parents above.

This is why all libuv tests are called `uv_*`, for example. Otherwise, it's impossible to determine which tests belong where, and further impossible to filter them out.

Qix-:
Not sure why Github is showing it as outdated but https://github.com/skypjack/uvw/pull/131#discussion_r222989926

test/CMakeLists.txt
skypjack: Oh, ok, didn't get the reason. Then let's use the same convention of `libuv`, thus `uvw_`. :+1:

Qix-: Sure :)

skypjack:
Ok. Let's see if travis gives us the go to merge. :wink:

Qix-:
Had one final touch I saw I could add. That should be it :)

Issue #132, opened 2018-10-06
uvw::CheckEvent is never fired.


Qix-:
The following test code hangs but no output occurs. A run through with lldb/gdb prove the callback is never hit.

```c++
#include <iostream>

#include "uvw.hpp"

using namespace std;

int main() {
	auto loop = uvw::Loop::getDefault();
	auto handle = loop->resource<uvw::CheckHandle>();

	handle->on<uvw::ErrorEvent>([](const auto &, auto &) {
		cerr << "ERROR" << endl;
	});

	handle->on<uvw::CheckEvent>([](const auto &, auto &) {
		cerr << "poll" << endl;
	});

	handle->start();

	loop->run();
}
```

Running with `NOWAIT`, as in the tests, correctly show `poll` once, but the tests don't check an actual loop.

```c++
loop->run<uvw::Loop::Mode::NOWAIT>();
```

I'll submit a failing test case.

Qix-:
When loop mode is default, `startCallback()` in `check.hpp` will never get hit, but since the handle is active the loop will run indefinitely.

Qix-:
Switching out the loop run with:

```c++
    for (int i = 0; i < 100; i++) {
        std::cerr << "loop " << i << std::endl;
        loop->run<uvw::Loop::Mode::ONCE>();
    }
```

Outputs `loop 0` and then hangs.

Qix-:
Actually, this appears to be an issue with libuv. The following hangs as well.

```c
#include <stdio.h>
#include "uv.h"

void on_check(uv_check_t *h) {
	puts("check");
}

int main(void) {
	uv_loop_t loop;
	uv_loop_init(&loop);

	uv_check_t ch;
	uv_check_init(&loop, &ch);
	uv_check_start(&ch, &on_check);

	uv_run(&loop, UV_RUN_DEFAULT);

	return 0;
}
```

I must be missing something with how checks work...

Qix-:
Yep. I guess this is a super weird case with libuv. It causes a deadlock - I might open a ticket over there about it.

- Checks are not fired until the polling is completed
- Polling occurs even if there are no descriptors to poll

Therefore, if the only active handle is a `uv_check_t`, the loop deadlocks immediately.

Not a bug in UVW, sorry about the noise (:

PR #133, opened 2018-10-06
Add failing check test

Qix-:
Ref #132.

Perhaps I'm missing something with how uvw is supposed to be used, but if that's the case then it's not entirely analogous to libuv.


Qix-:
The tests will hang until Travis kills them after a long while, by the way. You might want to cancel them yourself.

Qix-:
See https://github.com/skypjack/uvw/issues/132#issuecomment-427544384 for closure reason.

PR #155, opened 2019-07-17
Update request.hpp for cpp17, cpp20

wnsgml972:
This pull request contains the following changes:
* Change `std::result_of_t` to `std::invoke_result_t` because `std::result_of_t` is being deprecated in C++17.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,3 +21,4 @@ Miigon
 slyshykO
 bmagistro
 richardbmx
+wnsgml972
--- a/src/uvw/request.hpp
+++ b/src/uvw/request.hpp
@@ -27,20 +27,20 @@ protected:
         else { ptr->publish(E{}); }
     }
 
-    template<typename F, typename... Args>
-    auto invoke(F &&f, Args&&... args)
-    -> std::enable_if_t<not std::is_void<std::result_of_t<F(Args...)>>::value> {
-        auto err = std::forward<F>(f)(std::forward<Args>(args)...);
-        if(err) { Emitter<T>::publish(ErrorEvent{err}); }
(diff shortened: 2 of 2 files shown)


skypjack:
Hi, thanks for the PR.

A few notes:
* It doesn't follow coding style (actually it breaks it deliberately).
* CI returns an error, in particular appveyor.

I cannot merge this as-is. Do you plan to update it?

wnsgml972:
Thank you very much for your reply.

image

The future of `std :: result_of _t` is as follows. And I want to use it in std:cpp17 as well.

* So I implemented it using Preprocessor so that both std:cpp14 and std:cpp17 can operate.
* The working compiler added gcc, clang, and msvc with reference to CI configuration.

I have confirmed that my project now works normally on std:cpp17.
Thanks for reading!

skypjack:
To be honest, I wouldn't care much of maintaining the C++14 version.  
This is an open source library and it received few contributions nor much support so far, despite the number of users. Therefore I don't want to invest my time to maintain a C++14 version, mainly because I don't need it.  
I suggest to switch to C++17 and that's all. It will reduce also the codebase being it more expressive.

I'm out of home until the weekend.  
I'll edit the PR when I'm back and merge it.

wnsgml972:
Thank you for your suggestion.

I'll try to use it for my project and migrate it from std:cpp14 to std:cpp17.
Would it be OK to leave the PR again if it is finished?

Thank you for your feedback during your busy schedule.
Have a good weekend!

skypjack:
Working to a full migration towards C++17 and a restyling of the project. :+1: 
Be patient, it will take a while but I won't ignore the request. :wink:

PR #157, opened 2019-08-27
Add fromRaw function to loop

tversteeg:
I needed to add this function to be able to use a [luv](https://github.com/luvit/luv) loop in my code.

```c++
auto loop = uvw::Loop::fromRaw(luv_loop(...));
loop->run<uvw::Loop::Mode::NOWAIT>();
```

### DIFF
--- a/src/uvw/loop.hpp
+++ b/src/uvw/loop.hpp
@@ -201,6 +201,35 @@ public:
         return loop;
     }
 
+    /**
+     * @brief Wrap an existing loop libuv pointer.
+     *
+     * It may return an empty pointer in case of failure.<br>
+     * This function is a way to use external libraries that expose their own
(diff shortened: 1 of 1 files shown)


skypjack:
Wouldn't it be better something like this?
```
    static std::shared_ptr<Loop> create(uv_loop_t *ext) {
        auto ptr = std::unique_ptr<uv_loop_t, Deleter>{ext, [](uv_loop_t *){}};
        return std::shared_ptr<Loop>{new Loop{std::move(ptr)}};
    }
```

tversteeg:
I was considering this as well, but seeing that using the `create()` constructor would be the default recommended way and you wanting to discourage people to use raw C pointers I thought it would be better to create a different constructor for it.

PR #163, opened 2019-10-18
update typos and ambiguous phrases in README

lpmi-13:
I assumed the updates are closer to the intended meaning, but I'm happy to amend this pull request if I've misunderstood some things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ graceful C++ API. Currently, no `uv_*_t` data structure is actually exposed by
 the library.<br/>
 Note that `uvw` stays true to the API of `libuv` and it doesn't add anything to
 its interface. For the same reasons, users of the library must follow the same
-rules who are used to follow with `libuv`.<br/>
+rules which are used with `libuv`.<br/>
 As an example, a *handle* should be initialized before any other operation and
 closed once it is no longer in use.
 
@@ -82,7 +82,7 @@ int main() {
 
 ## Motivation
 
-The main reason for which `uvw` has been written is the fact that it does not
+The main reason for which `uvw` has been written is the fact that there does not
 exist a valid `libuv` wrapper in C++. That's all.
 
 # Build Instructions
@@ -102,7 +102,7 @@ documentation:
 
 Note that `libuv` is part of the dependencies of the project and it will be
 cloned by `cmake` (see below for further details).<br/>
-Because of that, users have not to install it to compile and execute the tests.
+Because of that, users don't have to install it to compile and execute the tests.
 
 ## Library
 
@@ -117,7 +117,7 @@ It's a matter of adding the following line at the top of a file:
 
 Then pass the proper `-I` argument to the compiler to add the `src` directory to
 the include paths.<br/>
-Note that users are demanded to correctly setup include directories and
+Note that users are required to correctly setup the include directories and
 libraries search paths for `libuv`.
 
 ## Versioning
@@ -172,7 +172,7 @@ reasons.
 ## Tests
 
 To compile and run the tests, `uvw` requires `libuv` and `googletest`.<br/>
-`cmake` will download and compile both the libraries before to compile anything
+`cmake` will download and compile both the libraries before compiling anything
 else.
 
 To build the tests:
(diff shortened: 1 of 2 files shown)


skypjack:
Thank you **very much**!! At a first glance there is only one point where you changed the meaning and therefore I'd ask to amend. I'll start a review for that. :+1:

README.md
```diff
@@ -117,7 +117,7 @@ It's a matter of adding the following line at the top of a file:
 
 Then pass the proper `-I` argument to the compiler to add the `src` directory to
 the include paths.<br/>
-Note that users are demanded to correctly setup include directories and
+Note that users are required to correctly setup, including directories and
```
skypjack: Here the meaning is changed. Users have to set up correctly the include directories and the libraries search paths. It's not _setting it up, include directories ..._, but _setting up include directories ..._. I'm not sure you got what I mean. Let me know if it's not clear.

README.md
```diff
@@ -207,7 +207,7 @@ For more details, please refer to the
 Initialization is usually performed under the hood and can be even passed over,
 as far as handles are created using the `Loop::resource` member function.<br/>
 On the other side, handles keep themselves alive until one explicitly closes
-them. Because of that, memory usage will grow up if users simply forget about a
+them. Because of that, memory usage will grow if users simply forget about a
```
skypjack: Just curious. Why is _up_ wrong here?

README.md
```diff
@@ -232,8 +232,8 @@ is enough, it's easy as doing this:
 auto loop = uvw::Loop::getDefault();
 ```
 
-Note that loop objects don't require to be closed explicitly, even if they offer
-the `close` member function in case an user wants to do that.<br/>
+Note that loop objects don't require being closed explicitly, even if they offer
+the `close` member function in case a user wants to do that.<br/>
```
skypjack: Shouldn't it be _an user_ rather than _a user_?

README.md
```diff
@@ -341,7 +341,7 @@ A connection object can be used later as an argument to the `erase` member
 function of the resource to remove the listener.<br/>
 There exists also the `clear` member function to drop all the listeners at once.
 
-Almost all the resources use to emit `ErrorEvent` events in case of errors.<br/>
+Almost all the resources use `ErrorEvent` to emit events in case of errors.<br/>
```
skypjack: `ErrorEvent` isn't a function, it's the emitted event(s). I think the meaning changed here.

skypjack:
Feel free also to add your nick to the [list of contributors](https://github.com/skypjack/uvw/blob/master/AUTHORS). It's in lexical order. :+1:

README.md
lpmi-13: you could use "up", though that's usually for people, like "the children grow up". It's not wrong, just possibly confusing. I'm happy to change it back if you would prefer the original.

README.md
lpmi-13: ah, got it. Will push an update.

README.md
lpmi-13: the a/an difference is usually in the sound, rather than the letter. So, for example, we'd probably use "an umbrella", cause it starts with a vowel sound, whereas something like "a university", cause it starts with more of a consonant sound ("yooniversity"). Here, user sounds like "yooser", so that's why it would normally go with "a".

README.md
lpmi-13: got it...will push an update.

README.md
skypjack: No, don't worry, I was just curious. You don't need to change it back. :wink:

README.md
skypjack: Got it. Thank you for the explanation. :+1:

skypjack:
Oh, have you 'push --force`d to update the branch? I suggest not to do that when you work on a PR because I missed the fact that you updated it and I can still squash all your commits on merge, so there are no real benefits.
Anyway, thank you very much. Do you want to add yourself as a contributor in the AUTHORS file before I merge this? Lexical order.

lpmi-13:
Oh that's weird.. does GitHub not notify you on force pushes? Just regular commits?

Sorry, I wouldn't normally rebase but figured there wasn't any risk of anyone doing work off that branch. Good to know re: notifications.

Anyway, have added my name to AUTHORS and added that to this PR (without force pushing haha).

skypjack:
Apparently, GH doesn't notify you. Just to be clear, the problem wasn't the rebase or the force push, it's just that I completely missed the fact that you updated it and it took way longer to merge it on `master`. :wink: 
Btw it's upstream, thank you very much for your help.

PR #171, opened 2020-02-28
Allow static build.

stefanofiorentino:
The aim of this PR is to give users a static build target `uvw_a`.
Users will be able to compile `uvw` as a dependency in their project.
In order to statically build the `uvw`, the cmake flag -DBUILD_STATIC_LIB=ON has to be added and the `uvw_a` has to be added to the project as a link target.


skypjack:
I'm pretty sure a lot of people will appreciate this. :+1:

PR #173, opened 2020-03-01
Changed FsReq::stasfs method name to statfs

sgiurgiu:
I believe it was a small typo in the method name, as uv does name it uv_fs_statfs and the method name in Linux is also statfs


skypjack:
Thanks for pointing this out.

Issue #196, opened 2020-05-16
BUILD_UVW_LIBS doesn't work on Windows (no symbols are exported)


amurzeau:
Hi,

When building UVW as a shared library (by setting BUILD_UVW_LIBS to ON) on Windows, a DLL is created but without any .lib file.
The issue is that nothing is exported using__declspec(dllexport).

I suggest to make the same thing as libuv, that is:
https://github.com/libuv/libuv/blob/36549815eed297c74448ff580b5c84713245ec5d/include/uv.h#L30

And putting the equivalent of UV_EXTERN to classes and functions.
This would allow to use -fvisibility=hidden with gcc too.

Are you willing to support that ?

Thanks :)

stefanofiorentino:
@amurzeau have you a working sample? 
I'll dive into this asap.

skypjack:
@amurzeau good point, the CI tests the flag with VS2019 but it doesn't really link to any library.

>Are you willing to support that ?

Sure, this makes perfectly sense. Did you have a try with it and have something to attach to a PR eventually?
This would speed up the whole thing a little. I'm short in time at the moment and it will take a while otherwise.

amurzeau:
Yes I've tried it and tested it by linking tests to uvw-shared instead of uvw-static.
I can come with a PR for an overview of the solution

skypjack:
That would be great. If you had code to use as a starting point, you could speed up the whole process a lot. Thanks!

amurzeau:
I see that in CMakeLists.txt in tests there is:
```cmake
$<$<TARGET_EXISTS:uvw::uvw>:uv::uv-static>
$<$<TARGET_EXISTS:uvw::uvw-static>:uvw::uvw-static>
```

Should'nt that be this instead ?:
```cmake
target_link_libraries(
         ${TEST_NAME}
         PRIVATE
             $<$<TARGET_EXISTS:uvw::uvw>:uvw::uvw>
-            $<$<TARGET_EXISTS:uvw::uvw>:uv::uv-static>
-            $<$<TARGET_EXISTS:uvw::uvw-static>:uvw::uvw-static>
+            $<$<TARGET_EXISTS:uv::uv-shared>:uv::uv-shared>
+            $<$<TARGET_EXISTS:uvw::uvw-shared>:uvw::uvw-shared>
             GTest::Main
             Threads::Threads
             ${LIBRT}
             ${WINSOCK2}
         )
```

amurzeau:
[...]

PR #197, opened 2020-05-17
WIP: shared libs: export classes with functions implemented in .cpp files

amurzeau:
When building a DLL on Windows or building a shared library with
`-fvisibility=hidden` on Unix, class with functions implemented in a cpp
file must be exported.

This mean `__declspec(dllexport)` on Windows, and
`__attribute__((visibility(default)))` on Unix.

Changes made:
 - `UVW_EXTERN` on all non templated classes.

 - Declare `UVW_EXTERN` in config.h.

 - Put `UVW_EXTERN` on all extern explicit template instantiations.

 - Put `UVW_EXTERN` on instantiation of `Flags<T>` from an exported template.
   MSVC was throwing a `C4251` warning for these, so I'm exported them too
   via an explicit template instantiation. This is not required for the code
   to link in fact.

 - Also add `UVW_MSVC_WARNING_PUSH_DISABLE_DLLINTERFACE()` for MSVC as it
   outputs many `C4251` warnings when an exported class uses a non exported
   template in the standard library. This means the user of the DLL must use
   similar enough compiler and flags so the C++ ABI is the same.
   There are these macro in all headers so the warning is still enabled for
   user code, if the user happen to use `__declspec(dllexport)` too.

 - Almost all exported classes derive from a template class. This doesn't
   trigger warning `C4275` because the base template class use the derived
   class in its template arguments. See last point here: https://docs.microsoft.com/en-us/cpp/cpp/general-rules-and-limitations?view=vs-2019

 - Change tests link libraries to use `uvw-shared` and `uv-shared` instead.
   So this will test that functions are really available from the DLL/.so.

 - Add CMake code to declare `USING_UVW_SHARED` or `BUILDING_UVW_SHARED` as
   appropriate.

 - Enable `CXX_VISIBILITY_HIDDEN` so only explicitly exported symbols are
   exported.

**Currently not working because of static variables in `Emitter<>::next_type()`
and `Emitter<>::event_type<>()`.**

Fixes: #196

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -74,8 +74,20 @@ add_uvw_library(uvw-static)
 add_library(uvw-shared SHARED)
 add_library(uvw::uvw-shared ALIAS uvw-shared)
 target_link_libraries(uvw-shared PUBLIC uv::uv-shared)
-set_target_properties(uvw-shared PROPERTIES VERSION ${PROJECT_VERSION} SOVERSION ${UVW_VERSION_MAJOR} EXCLUDE_FROM_DEFAULT_BUILD 1)
+set_target_properties(uvw-shared PROPERTIES
+	VERSION ${PROJECT_VERSION}
+	SOVERSION ${UVW_VERSION_MAJOR}
+	EXCLUDE_FROM_DEFAULT_BUILD 1
+	C_VISIBILITY_PRESET hidden
+	CXX_VISIBILITY_PRESET hidden
+)
 add_uvw_library(uvw-shared)
+target_compile_definitions(uvw-shared
+  INTERFACE
+    USING_UVW_SHARED=1
+  PRIVATE
+    BUILDING_UVW_SHARED=1
+)
 
 #
 # Install targets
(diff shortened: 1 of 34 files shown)


stefanofiorentino:
> Currently not working because of static variables in Emitter<>::next_type() and Emitter<>::event_type<>().

@amurzeau I have spent some time trying avoiding the monostate pattern causing your testing errors. 
Just for the readers, this kind of issue arises because with dynamic linking your monostate's statics are two, one in the DLL and one in the executable, so your events don't match any event in the DLL and viceversa.
An available solution to this is to move events creation to compile-time (generating a lot of different types of `Emitter<>::Handler`s). Regrettably this has several drawbacks: majors are the cardinality of the set of different Handlers and the need of the /bigobj flag under MSVC. I think I'll give up on this for a while waiting for more inspiration (i.e. for a better solution). 
Meanwhile we should use the static linking (imho best linking method anyway).

@skypjack I don't know if you want to close this PR as `won't do`. Choice is yours.

skypjack:
I'm implementing yet another approach that makes the event identification somewhat portable across boundaries.
I've not much bandwidth, so it may take a while, but still. We can close the PR, we've an issue anyway so as not to forget to work on this. 😉

Issue #201, opened 2020-05-28
uvw OpenBSD support


elindsey:
We got a report of one of our projects that uses uvw failing to build on OpenBSD - https://github.com/DNS-OARC/flamethrower/issues/44

Looking into it, uvw _almost_ works on OpenBSD, but there are a few small things preventing it. 🙂 

From a clean OpenBSD 6.7 install with uvw cloned into /tmp:
```bash
openbsd# pkg_add libuv
openbsd# pwd
/tmp/uvw
openbsd# ls
test.cpp uvw      uvw.hpp
openbsd# cat test.cpp                                                                                                                                                                                  
#include "uvw.hpp"

int main() {

}

openbsd# clang++ -std=c++17 test.cpp -I/usr/local/include 
In file included from test.cpp:1:
In file included from ./uvw.hpp:1:
In file included from ./uvw/async.hpp:7:
./uvw/handle.hpp:267:22: error: expected member name or ';' after declaration specifiers
    OSFileDescriptor fileno() const {
    ~~~~~~~~~~~~~~~~ ^
/usr/include/stdio.h:406:20: note: expanded from macro 'fileno'
#define fileno(p)       (!__isthreaded ? __sfileno(p) : (fileno)(p))
                         ^
In file included from test.cpp:1:
In file included from ./uvw.hpp:1:
In file included from ./uvw/async.hpp:7:
./uvw/handle.hpp:267:22: error: expected ')'
/usr/include/stdio.h:406:20: note: expanded from macro 'fileno'
#define fileno(p)       (!__isthreaded ? __sfileno(p) : (fileno)(p))
                         ^
./uvw/handle.hpp:267:22: note: to match this '('
/usr/include/stdio.h:406:19: note: expanded from macro 'fileno'
#define fileno(p)       (!__isthreaded ? __sfileno(p) : (fileno)(p))
                        ^
2 errors generated.
```

The main issue is that on OpenBSD, `fileno` is a macro and will clash with the name of this handle method. The work is small, but it would be a breaking API change to rename `Handle::fileno`. 

There are four things that would need to be done for OpenBSD support:
1. Rename the `Handle::fileno` method
2. Small cmake fixes - don't try to link librt on BSD; adjust error levels so implicit integer conversion in gtest doesn't fail the build
3. ifdef out `sendfile` functionality (not present on OpenBSD so those tests fail)
4. Investigate the fs_events tests (those stall on OpenBSD; I haven't looked into OS support)

[...]

PR #204, opened 2020-05-30
add openbsd support

elindsey:
Renames fileno() so it doesn't clash with OpenBSD's macro expansion.

Makes minor changes to build steps so the test suite compiles and runs
on OpenBSD.

See https://github.com/skypjack/uvw/issues/201 for full context

### DIFF
--- a/src/uvw/handle.hpp
+++ b/src/uvw/handle.hpp
@@ -264,7 +264,7 @@ public:
      * @return The file descriptor attached to the hande or a negative value in
      * case of errors.
      */
-    OSFileDescriptor fileno() const {
+    OSFileDescriptor fd() const {
         uv_os_fd_t fd;
         uv_fileno(this->template get<uv_handle_t>(), &fd);
         return fd;
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -29,7 +29,6 @@ else()
 
     target_compile_features(gtest PUBLIC cxx_std_17)
     target_compile_features(gtest_main PUBLIC cxx_std_17)
-    target_compile_options(gtest PRIVATE -Wno-error)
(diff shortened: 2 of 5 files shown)


elindsey:
I'll fix those Windows CI failures - forgot msvc takes different flags.

elindsey:
I'm going to drop the no-error override and see if I can fix upstream gtest instead.

skypjack:
Ok. Thanks for the PR. I'll be taking a look to this one for updates. :+1:

elindsey:
Okay, so with this patch projects that use uvw can successfully build and run on OpenBSD (have confirmed with our project flamethrower). Additionally, uvw's test suite builds and runs with one test failure on OpenBSD.

There is an upstream patch to fix gtest's compilation failures on OpenBSD (https://github.com/google/googletest/pull/2872), but I don't know how long that will take to merge, so I've disabled treating warnings as errors specifically for gtest.

The remaining test failure is a libuv pthread_cond_wait segfault in the fs_event test. Unfortunately I'm about at my limit right now, so don't know when or if I'll be able to look into that one. I think the patch is valuable as-is since it unblocks most real use cases, but up to you. 🙂

skypjack:
I haven't looked in the documentation because I'm a bit busy at the moment but I dare to ask in case you know that already. Does GH Action support OpenBSD as a platform? I guess no.

elindsey:
Unfortunately it doesn't, run targets are very limited - macOS Catalina, Windows Server, and Ubuntu.

skypjack:
Well, I was waiting for a release of `libuv` to merge this. However, it's taking longer than expected.
I'll cut a patch release in a few days, so that you can proceed on your side. Sorry for being late.

Issue #203, opened 2020-05-30
Lost functionalities on Write


nicolasnoble:
There are two functionalities I am currently missing from using raw libuv.

The first one is the ability to issue a write request over multiple buffers. Right now the internal `WriteReq` object is hardcoded to invoke `uv_write` with only a single buffer. This is mostly relevant with the second missing functionality:

The second one is that there is no distinction between write requests when the `WriteEvent` is being fired. The `write` method returns a void, instead of exposing the `WriteReq`, and then it delegates the `WriteEvent` to its `StreamHandle`.

Ideally, one ought to be able to:

 - pass more than one buffer to `write`, and
 - receive a `WriteReq` as a response, that is then referenced into the `WriteEvent` being fired later.

The rationale is that depending on what sort of data being sent, a packet being sent may be for example constructed with a fixed header, a fixed footer, and a variable body for example, or a different combination thereof, this being the basic scatter-gather principle. Being able to manage the memory resource for the whole reconstructed write request if a fairly useful feature, which is currently missing.

nicolasnoble:
The first issue seems to be doable using another overload of the `write` method. Memory management in this case is fully on the caller's, and the API doesn't take any ownership for any of the buffers. 

An easy solution that doesn't break API for the second part would be to make `write` return a token of type `uintptr_t`, which will be guaranteed unique across all pending write requests. It'd also be passed into the `WriteEvent` structure.

The value of the token can be a simple cast from the inner write request pointer.

I can probably come up with a pull request for these tomorrow.

nicolasnoble:
So the real functionality being lost about the write request, is that uvw currently is the sole responsible for managing these resources. This deprives the user of being able to reuse these objets, in case they want to manage them themselves, thus creating unnecessary overhead.

Ultimately, it'd be ideal if the user can:

 - Derive an object from a `uvw::WriteReq`;
 - Optionally pass one of these to any of the `write` calls - if none is provided, then the library should be responsible for creating a generic one and deleting it afterward; and
 - Get the object back into the `WriteEvent` structure.

skypjack:
You can already create requests from a loop actually. Internally, this is what happens.
Also, requests already announce themselves when they _emit_ the `WriteEvent`, similarly to that handles do.
So, if I get it right, all you need is to make requests available rather than an implementation detail, so that you can create them and pass a stream to use to them. Is it?
[...]

PR #205, opened 2020-05-30
Adding tokens to write events.

nicolasnoble:
This addresses one part of #203.

### DIFF
--- a/src/uvw/request.hpp
+++ b/src/uvw/request.hpp
@@ -28,6 +28,13 @@ protected:
         else { ptr->publish(E{}); }
     }
 
+    template<typename E>
+    static void tokenCallback(U *req, int status) {
+        auto ptr = reserve(req);
+        if(status) { ptr->publish(ErrorEvent{status}); }
+        else { ptr->publish(E{reinterpret_cast<uintptr_t>(req)}); }
+    }
+
     template<typename F, typename... Args>
     auto invoke(F &&f, Args&&... args) {
         if constexpr(std::is_void_v<std::invoke_result_t<F, Args...>>) {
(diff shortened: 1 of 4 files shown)


nicolasnoble:
This still needs the multiple buffers code.

skypjack:
Also, `std::uintptr_t` isn't that portable. I'd rather get rid of it to avoid portability issues.

skypjack:
Moreover, I don't see exactly what the role of the _token_ is.
If all you need is to get a write request along with the write event, the former can be attached to the latter. Since handles in`uvw` emit _on themselves_, you're also guaranteed that a given request is what you expect it to be in a callback and not one randomly received from another handle.

nicolasnoble:
> If all you need is to get a write request along with the write event, the former can be attached to the latter.

That's basically what this is, yes. This tentatively hides the concept of write request out to the user, because that's the only need there can be for them. Getting a `WriteReq` might be more confusing, but that's okay too, since that's what libuv actually does. I'm fine either way personally.

> Since handles in`uvw` emit _on themselves_, you're also guaranteed that a given request is what you expect it to be in a callback and not one randomly received from another handle.

Which is fine when you're only have one outstanding write for a simple protocol, but if you have a multiplexing protocol with substreams, such as the likes of http/2, then you need a multiplexing/demultiplexing layer that lets you know which substream got its event fired. This is a solution for this. You can then very easily write something along this pseudo-c++:

```C
class Http2 {
  std::map<uvw::StreamHandle::TokenType, http2::SubStream*> m_outstandingWrites;

  void initEvents() {
    m_tcp->on<uvw::WriteEvent>([this](const uvw::WriteEvent &event, uvw::TCPHandle &) {
      m_outstandingWrites[event.token]->publish(Http2WriteEvent{});
      m_outstandingWrites.erase(event.token);
    }
  }
  uvw::TCPHandle m_tcp;

  class SubStream : public Publisher {
    void write(const Payload &payload) {
      auto token = m_parent->m_tcp->write(payload.bytes(), payload.length());
      m_outstandingWrites.insert(token, this);      
    }
  };
};
```

nicolasnoble:
And the more I'm thinking about it, the more I realize this isn't even going far enough. Closing this one and updating the associated bug.

PR #207, opened 2020-06-17
Cross-platform support to shared libs

### LINKED ISSUE #196: BUILD_UVW_LIBS doesn't work on Windows (no symbols are exported)

stefanofiorentino:
closes #196


stefanofiorentino:
This is intended for comparing against the current HEAD.
NOTE: still contains minor flaws about FakeEvent/Emitter test classses.

stefanofiorentino:
@skypjack this is the best solution I can find to have cross-platform support to shared libs. 
In my opinion the question rose by @amurzeau is right, do we want to fully support shared libs too?
From my point of view, the price is simplicity and readability.

skypjack:
>From my point of view, the price is simplicity and readability.

Do you mean not supporting them? Because this change isn't exactly simple nor easy to read. :smile:

stefanofiorentino:
I mean we can support only static linking (not the shared one).
The simplicity and readability will be impacted a lot by this mod.

skypjack:
>The simplicity and readability will be impacted a lot by this mod.

Indeed, it doesn't seem to be something simple to maintain.

stefanofiorentino:
I'll close this as per discussion.

PR #215, opened 2020-07-28
fix undefined behavior of Loop::walk

### LINKED ISSUE #214: conversion from void* to the pointer of the base class is not a good idea
https://github.com/skypjack/uvw/blob/a10844d4bfd1e31bd062af9aae65add0e70cc136/src/uvw/resource.hpp#L44

https://github.com/skypjack/uvw/blob/a10844d4bfd1e31bd062af9aae65add0e70cc136/src/uvw/loop.cpp#L113

basehandle::type() in uv_walk which is in a timer callback still causes an undefined behavior.

On windows, BaseHandleWalk test failed.
图片

Related issue:
https://github.com/skypjack/uvw/issues/212

Reference:
https://stackoverflow.com/a/5445220

DuckVador:
Close #214

Passed all tests with ASAN and UBSAN.

### DIFF
--- a/src/uvw/async.cpp
+++ b/src/uvw/async.cpp
@@ -9,7 +9,7 @@ namespace uvw {
 
 
 UVW_INLINE void AsyncHandle::sendCallback(uv_async_t *handle) {
-    AsyncHandle &async = *(static_cast<AsyncHandle *>(handle->data));
+    AsyncHandle &async = downcast(handle->data);
     async.publish(AsyncEvent{});
 }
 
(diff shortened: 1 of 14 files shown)


skypjack:
This doesn't convince me to be honest. There are some things that can cause problems in the short term. Let me look into it.

skypjack:
Please, see the discussion on #214 and branch `experimental` for further details. Thanks.

DuckVador:
https://github.com/skypjack/uvw/commit/0b329720bfd938af22a59a47d72aa1ef7094de58 is better implementation options, close it.

Issue #223, opened 2020-10-07
[Feature request] A customizable allocate memory function


DuckVador:
`UVW` provides a `uv_alloc_cb` to allocate memory with `new`, but this is inconvenient when I want to manage the buffer more complex or use `malloc` to allocate memory.

stefanofiorentino:
Maybe somehow related to #182? (i.e. is it a generalization of that issue?)

DuckVador:
#182 is about write operations. This issue is more about read operations.

`uvw` allocate memory function:
https://github.com/skypjack/uvw/blob/77af4a3fc4d932a52652807506fc50d0e58e875c/src/uvw/handle.hpp#L40-L43

`uvw` read callback:
https://github.com/skypjack/uvw/blob/77af4a3fc4d932a52652807506fc50d0e58e875c/src/uvw/stream.h#L131-L134

`uvw` `DataEvent` (`UDPDataEvent`): 
https://github.com/skypjack/uvw/blob/77af4a3fc4d932a52652807506fc50d0e58e875c/src/uvw/stream.h#L64-L67
https://github.com/skypjack/uvw/blob/77af4a3fc4d932a52652807506fc50d0e58e875c/src/uvw/udp.h#L32-L35

Since the memory allocation function uses `operator new`, the implementation of `DataEvent` assumes that unique_ptr uses the default deleter. Because these are not customizable, I can't control the process of allocating memory even if I have a chance to avoid `operator new`.

skypjack:
How do you envision setting it up? Because doing it on a per-handle basis seems a little annoying.

DuckVador:
I think the default behavior is enough for most cases. Sometimes when the user wants to buffer reuse such as using a C library which buffer pointer can be then `realloc` etc, they must define an `allocator` which provides a deleter and an allocator.
`uvw` default `allocator` could be a simple trait of `new []` and `std::default_delete<char[]>`.

skypjack:
Do you mind creating a PR for that? Not even a completely refined one, but having something that shows how you would intervene would be great if possible.

DuckVador:
OK, I will make a basic implementation of my idea.

DuckVador:
@skypjack I have created a PR of the custom allocator. Please have a look.

skypjack:
Hi, sorry for being late. I had some personal problems over the weekend and I didn't even turn on my pc.
Thank you for the PR. I'm tryng to keep up with all the things I left behind. I'll look into it soon. 👍

PR #225, opened 2020-10-09
Custom allocator experiment

DuckVador:
A basic implementation of #223 

The custom allocator test is in `test/main.cpp`.

### DIFF
--- a/src/uvw/handle.hpp
+++ b/src/uvw/handle.hpp
@@ -21,6 +21,27 @@ namespace uvw {
 struct CloseEvent {};
 
 
+class Allocator{
+public:
+    virtual uv_buf_t allocate(std::size_t suggested)
+    {
+        auto size = static_cast<unsigned int>(suggested);
+        return uv_buf_init(new char[size], size);
+    }
+
+    virtual void deallocate(char* ptr)
+    {
+        delete [] ptr;
+    }
+
+    virtual ~Allocator()= default;;
+
+    static Allocator* getDefaultAllocator(){
+        static Allocator defaultAllocator;
+        return &defaultAllocator;
+    }
+};
+
 /**
  * @brief Handle base class.
  *
@@ -29,6 +50,8 @@ struct CloseEvent {};
 template<typename T, typename U>
 class Handle: public Resource<T, U> {
 protected:
+    Allocator* allocatorPtr = Allocator::getDefaultAllocator();
+
 	static void closeCallback(uv_handle_t *handle) {
 		Handle<T, U> &ref = *(static_cast<T*>(handle->data));
 		auto ptr = ref.shared_from_this();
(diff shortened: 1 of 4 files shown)


skypjack:
@stefanofiorentino what do you think about this? 🙂

stefanofiorentino:
I think `allocation` is always an interesting topic. Will review this asap

stefanofiorentino:
Frankly, I'm a bit confused about what's the goal over here.
@DuckVador can you briefly show a "real" use case of this?
The `test/main.cpp` doesn't let me think of a useful use case, sorry for this.

DuckVador:
@stefanofiorentino 
libuv [doc](http://docs.libuv.org/en/v1.x/handle.html) gives a very brief introduction with custom allocation。 
> As an example, applications with custom allocation schemes such as using freelists, allocation pools or slab based allocators may decide to use a different size which matches the memory chunks they already have.

I think a general technique for memory allocation optimization is allocation pools. 
This [answer](https://stackoverflow.com/a/28780588/5846225) give a C implementation of the original libuv API.

stefanofiorentino:
Got it, as me, you don't like memory increasing and decreasing over time.
Now, I'm just wondering if we can arrange a better solution to assign the allocator at runtime (i.e. I don't love the `setAllocator` call). Let me think about this new issue while commuting this evening.

skypjack:
As far as I can see, we can assign allocators on a per-handle basis at runtime, on a per-type basis at runtime or on a per-type basis at compile-time. Probably also on a per-handle basis at compile-time, even though this will literally duplicate the number of symbols.

stefanofiorentino:
@DuckVador I got a further question for you.
Are you ok with just the custom allocation for read buffer, or your final target is to have a memory pool for every allocation in the lib? This question arose to me highlighting how compliant we have to stay to the STL stuff (and the trend to be as much as possible compile-time).

DuckVador:
@stefanofiorentino For me, my goal is to customize the read buffer, a custom allocator at runtime (like OpenCV), or at compile time (like STL) is acceptable.

stefanofiorentino:
As I would stick to the standard stuff as much as possible, I have to pass back the baton.
@skypjack what do you think? Would you accept a refinement of this proposal or you would like to introduce something different.
Waiting for `std::pmr` is not an option, I sadly think.

skypjack:
Unfortunately, `std::pmr` is literally unusable in the vast majority of cases, especially in those where you would use `libuv`.
Personally, I think a compile-time setting can be easier to work with in the long term rather than a per-handle solution like this one. Though, I'm open to feedback.

DuckVador:
Waiting for a compile-time version of custom allocator. Close this PR. :)

stefanofiorentino:
Let's continue investigating within next few weeks as it can be useful for several application.

PR #263, opened 2022-04-07
UVW v3

skypjack:
A breaking change to set a tipping point for a new version of uvw, an updated coding style (with a more _standard-ish_ snake style) and huge performance improvements during compilation.
Feedback are highly appreciated. 🙂


stefanofiorentino:
@skypjack do you need any help to get the pipeline back to green too?

skypjack:
Oh, well, put aside some sporadic failures mainly due to the poll tests, it depends on the reduced test coverage and I'm trying to increase it. Any help is appreciated ofc. 🙂

sthagen:
I really like the reduced compile time and find the `multi_word_style` much better than `multiWordStyle` for readers. 

(not) just joking: 
>Optimal would be to have a programming language that allows spaces in identifiers and ideal if we had enough words to only ever use single word identifiers ... but these are topics left as exercises for later generations (and languages).

sthagen (dismissed):

skypjack:
@sthagen have you tried this branch already?
I'm observing a huge improvement in terms of compilation time and I'm curious to know if it's a common thing.
Also, feel free to drop comments, suggestions and complaints if you want. Any contribution is appreciated! 🙂

sthagen:
@skypjack I did not, but I hope to do it on the weekend (... there are traces of such "plans of mine" here on github where a day became a month ... 🤷🏽‍♂️)
I read through the hunderd or so files in the change set and in theory it really looks good. Practice may decide differently on my machine, but I am confident and happy your machines already show the practical gains) 🚀 

Sometimes when many functions change their return type from `void` to `int` (like in this PR) change sets risk to become a liability. But, typically in wrappers / proxies like this library, this only indicates the beneficial move from a hiding facade pattern to an open proxy that propagates the state info from the lower layers back to the caller instead of providing riddles to the callers :+1:

skypjack:
Yeah, the idea is to make it more _transparent_ and only add the work around memory and lifetime management on top of `libuv`. It doesn't make much sense to _hide_ to the final user the error codes or any other thing, as if there wasn't `libuv` under the hood. 😅

stefanofiorentino (approved):

stefanofiorentino:
@skypjack it fails tests on my env (WSL2 ubuntu-20.04):
```
[ctest] ../test/uvw/loop.cpp:86: Failure
[ctest] Expected equality of these values:
[ctest]   count
[ctest]     Which is: 11
[ctest]   12u
[ctest]     Which is: 12

**Comment by skypjack** [comment]
Mmm any chance you can reproduce it? I don't know what's going on there otherwise. 😞

**Comment by skypjack** [comment]
> Mmm any chance you can reproduce it? I don't know what's going on there otherwise. 😞

@stefanofiorentino ping 🙂 
I'd like to merge this one upstream but I can't reproduce nor fix the failure you mentioned.
Any news about it?

**Comment by fiorentino-at-adesso** [comment]
> > Mmm any chance you can reproduce it? I don't know what's going on there otherwise. 😞
> 
> @stefanofiorentino ping 🙂 I'd like to merge this one upstream but I can't reproduce nor fix the failure you mentioned. Any news about it?

Some (little) investigations yield this: 
- the failing test is Loop.walk
- failing with USE_ASAN and USE_UBSAN, too

I'm wondering, why we don't have USE_*SAN as github workflows?

**Comment by stefanofiorentino** [comment]
Some (little) investigations yield this:

- the failing test is Loop.walk
- by enabling USE_ASAN the "Lopp" tests pass, but "616 byte(s) leaked in 5 allocation(s)".

I'm wondering, why we don't have USE_*SAN as github workflows?

**Comment by skypjack** [comment]
> I'm wondering, why we don't have USE_*SAN as github workflows?

Testing a workflow for that on this branch. 👍

**Comment by skypjack** [comment]
Cool @stefanofiorentino at least I can reproduce the error on the CI. Good catch. 👍

**Comment by stefanofiorentino** [comment]
@skypjack launching them 1-by-1 with --gtest_filter only `Loop.Close` causes the leak.

**Comment by stefanofiorentino** [comment]
Probably the uv_loop_close returns UV_EBUSY. I've to stop investigations for a while.

**Comment by skypjack** [comment]
Oh, wait, I suspect the test is wrong actually. We're closing the handle after the loop. 🤦‍♂️ 
Gotta go now but I'll give it a try during the weekend if possible. Thanks for checking it. ❤️

**Comment by stefanofiorentino** [comment]
I see the workflows here are green, but my local build (after a git force-clean) still complains: 

> [test](v3)$ ./loop --gtest_list_tests | grep "^  " | sed 's/^  //g' | xargs -r -n1 -I '{}' bash -c "./loop --gtest_filter=*{} 2&>1 > /dev/null || echo {} test is KO"
> Walk test is KO

**Comment by skypjack** [comment]
What error do you get @stefanofiorentino `Loop.walk`, a wrong counter?

**Comment by stefanofiorentino** [comment]
> What error do you get @stefanofiorentino `Loop.walk`, a wrong counter?

Yes, my cmake conf is 
```
    "cmake.configureArgs": [
        "-DBUILD_TESTING:BOOL=TRUE"
    ]
```
and the error is still 
```
[ctest] ../test/uvw/loop.cpp:88: Failure
[ctest] Expected equality of these values:
[ctest]   count
[ctest]     Which is: 11
[ctest]   12u
[ctest]     Which is: 12
```

**Comment by skypjack** [comment]
Yeah @stefanofiorentino that's this line:
```cpp
loop->resource<uvw::tty_handle>(0, true);
```
If you run it from the command line, it causes your error. If you can confirm it, we can also drop the line from the test. 👍

stefanofiorentino:
> Yeah @stefanofiorentino that's this line:
> 
> ```c++
> loop->resource<uvw::tty_handle>(0, true);
> ```
> 
> If you run it from the command line, it causes your error. If you can confirm it, we can also drop the line from the test. 👍

Ok, as `setsid sh -c "./test/loop"` is working for me.
I'm going to approve this ;)

stefanofiorentino (approved):

skypjack:
Any other concerns? I'd make this the upstream version. It has indecently faster compile times too.

stefanofiorentino:
@skypjack we should definitely celebrate this day.

skypjack:
Indeed, it's a huge improvement! Compilation time on the CI is already halved too. Very impressive.

PR #266, opened 2022-06-12
Add static lib when building with system's static libuv

darktohka:
When building with `-DFETCH_LIBUV=OFF`, `-DFIND_LIBUV=ON` and `-DBUILD_UVW_LIBS=ON` and a static build of libuv, the build fails with the following error:

```
-- Checking for module 'libuv>=1.44.1'
--   Found libuv, version 1.44.1
-- libuv 1.44.1 found via pkg-config
-- Configuring done
CMake Error at src/CMakeLists.txt:76 (target_link_libraries):
-- Generating done
  Target "uvw" links to:

    uv::uv-static


  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

This can be fixed by adding the uv-static target when building uvw as a static library.


skypjack:
LGTM. @stefanofiorentino ?
@darktohka can you point your PR to the v3 branch, please? It will replace `master` as a whole soon. Thanks.

stefanofiorentino:
ciao @darktohka, [this](https://github.com/skypjack/uvw/pull/266/commits/f571fc3cfd596e4bdeac609938d923ef2f79464c#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR121) line is not totally convincing me. Distribution packages does usually not provide the static version of libs. Have you a custom environment? (aka did you manually install libuv?)

darktohka:
Yes, I've compiled and installed libuv as a static library. pkgconfig can detect the static variant as well.

darktohka:
How about this?

We can instead search for libuv-static through pkgconfig, which is guaranteed to be a static library.

0EVSG:
Stumbled upon this because I just created a local FreeBSD package of uvw for testing:

1. It's not uncommon for system packages to provide both static and dynamic libraries, see e.g. [ubuntu](https://packages.ubuntu.com/jammy/amd64/libuv1-dev/filelist) or [FreeBSD](https://www.freshports.org/devel/libuv/). Names may differ, like `libuv.a`, `libuv_a.a` and PkgConfig names also (`libuv`, `libuv-static`).
2. CMake `pkg_check_modules` already [has](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) facilities for detecting both static and shared libraries in one go.
3. From a packagers POV, uvw should be able to create both static and shared libraries in the same build (with `FETCH_LIBUV` disabled). Header files certainly installed too, for header-only consumers.

skypjack:
> uvw should be able to create both static and shared libraries in the same build

Do you mean from the same build tree or from the same build command? As in _start the build and return with both the libraries compiled at once_?

0EVSG:
> Do you mean from the same build tree or from the same build command? As in start the build and return with both the libraries compiled at once?

Exactly. As in setup _one_ build directory with options set for what to build - headers, dynamic lib, static lib, or any combination of them. Fail right there in case the corresponding libuv libraries are not found. Then just build and install, usually in two steps, using make or ninja. I may have overlooked something, but from the CMakeLists.txt it looked like dynamic xor static to me.

Unfortunately I can't dive more into this, just learned that libuv only has timers with ms granularity, which doesn't cut it for me.
Good luck anyway!

stefanofiorentino:
@darktohka can we close this? 
@skypjack I think it's outdated at this point, right?

skypjack:
Yeah, definitely. We can close this one and open a new PR from current `master` eventually if there is something else to do.

stefanofiorentino:
@darktohka feel free to rebase and re-open it if you're still in the need of this modification. 
We improved this build setup lately, so maybe you can give it a try.

PR #269, opened 2023-01-05
shared object versioning (mirroring the behaviour of autotools)

### LINKED ISSUE #268: The library targets are not versioned.
Working on the Yocto Project [meta-layer](https://github.com/stefanofiorentino/meta-uvw.git) to natively list `uvw` in the OpenEmbedded Layer index and this [sample application](https://github.com/stefanofiorentino/uvw_static_lib_client.git) (the application delivered by `uvw` inside the test folder, brought in a separated repo), I got the error that the libraries provided (and installed) by this CMake-based project are not versioned (best-practice is that the libuvw.so lib is a symlink to the versioned "concrete" libuvw.so.2.11.2).
I'll work on it.

stefanofiorentino:
Closes #268

Signed-off-by: Fiorentino, Stefano <stefano.fiorentino@adesso.ch>


CMakeLists.txt
```diff
@@ -187,8 +187,23 @@ install(
 # Install targets
 #
 
-install(EXPORT uvwConfig NAMESPACE uvw:: DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/uvw)
-install(TARGETS uvw EXPORT uvwConfig ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+set_target_properties(
+    uvw PROPERTIES
+    VERSION ${UVW_VERSION_MAJOR}.0.0
```
skypjack, line 192: Why not `UVW_VERSION_MINOR` and `UVW_VERSION_PATCH`? What am I missing?

stefanofiorentino: I've just stayed true with libuv [style](https://github.com/libuv/libuv/blob/v1.x/CMakeLists.txt#L725): 
image
That's clear we can add more details with minor and patch, do you think is better to do so?

stefanofiorentino: It's a sort of common practice with autotools, within MAJOR' versions no APIs are broken.. and so on..
but you know what, let's add all info we already have for free in the version property.
I'll add my commit asap.

skypjack (approved):

PR #273, opened 2023-02-20
Include cstdint where uintX_t is used

pemensik:
Use namespaced version of stdint.h header, which defines int32_t, uint32_t and similar types.

This fixes build error when building flamethrower package with recent release of uvw.
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2171490


stefanofiorentino:
I've restarted failing runs. See you later.

skypjack (approved):

PR #281, opened 2023-05-03
Fix compile errors caused by ambiguous naming of poll_event in poll_handle

ReimuNotMoe:
The line 
```
using poll_event = details::uvw_poll_event;
```

created an ambiguous alias with the same name as the  `struct poll_event`.

And `poll_handle` should be inheriting `handle<poll_handle, uv_poll_t, poll_event>` instead of `handle<poll_handle, uv_poll_t, details::uvw_poll_event>` if I'm understanding correctly. Otherwise `on<uvw::poll_event>()` doesn't work.

### DIFF
--- a/src/uvw/poll.cpp
+++ b/src/uvw/poll.cpp
@@ -20,7 +20,7 @@ UVW_INLINE void poll_handle::start_callback(uv_poll_t *hndl, int status, int eve
     if(poll_handle &poll = *(static_cast<poll_handle *>(hndl->data)); status) {
         poll.publish(error_event{status});
     } else {
-        poll.publish(poll_event{poll_event(events)});
+        poll.publish(poll_event{details::uvw_poll_event(events)});
     }
 }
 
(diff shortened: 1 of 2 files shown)


skypjack:
You're definitely right on `handle<poll_handle, uv_poll_t, details::uvw_poll_event>`. 👍 
On the other hand, I'd rather change the using declaration to 🤔 `poll_event_flags` (we used a similar pattern somewhere else too) and use it in the class.
I don't like using `details::xxx` names in an API, it's odd and somewhat risky.

Do you mind updating the PR?

ReimuNotMoe:
> You're definitely right on `handle<poll_handle, uv_poll_t, details::uvw_poll_event>`. +1 On the other hand, I'd rather change the using declaration to thinking `poll_event_flags` (we used a similar pattern somewhere else too) and use it in the class. I don't like using `details::xxx` names in an API, it's odd and somewhat risky.
> 
> Do you mind updating the PR?

Sure.

skypjack (approved):

Issue #289, opened 2023-05-29
Provide a way to specify buffer size


ReimuNotMoe:
Hello. I'm trying to optimize the performance of uvw and libuv on low-ram (<64MiB) embedded platforms. Currently, all buffer allocations for stream handles are done by the `details::common_alloc_callback` function, and it just allocates as much ram as libuv suggested (64KiB). This is too big and undesirable on almost all embedded platforms. I wonder if it's possible to enable the user to specify the buffer size, or even better, use a user supplied allocation callback? Thanks.

skypjack:
Uhm, from what I see, we use it with `uv_read_start` and `uv_udp_recv_start`, that is, `stream_handle<...>::read` and `udp_handle::recv`.
The easiest thing to do is to accept a callback at the callsite probably. Passing around the buffer size is also possible but it should be an auto template argument to use to specialize `common_alloc_callback`. It doesn't worth it imho.
Any other suggestions on how to implement it otherwise?

ReimuNotMoe:
I can't think of other reasonable ways. Still I think allowing the user to set a callback is the best option, since this would allow the user to do further optimizations, such as using a memory pool. The next thing to worry about is how to provide the API, I'm fine with a new member function, and something like `on<uvw::alloc_event>()` could also work but it would introduce extra overhead (shared_ptr+function) and it will be the only callback that doesn't have uvw handle as the first argument...

ReimuNotMoe:
Yeah, I think so.

mfrg:
Also interested in this from a recvmmsg support perspective.  We need to be able to provide a buffer that is a multiple of the max datagram size for recvmmsg to work correctly.

Issue #267, opened 2022-08-05
I need the way to set buffer when I read from stream


hihig2001:
In stream like things, I need to collect several data events to make full packet.
but it seems there are lack of interface I can give my buffer data(aka. char[]) when I read from stream.

Because uvw stands for thin wrapper of libuv, I think it should be possible.

Current implementation allocates by uvw in
`Handle::allocCallback`

It seems when I write I have controls of buffer but when I read it seems not. 

If i can set this kind of call back to streams I could control it.
`auto alloc_callback(Handle<T, U>& handle, std::size_t suggested) -> char*`

I don't want allocation happens in every read operation.

skypjack:
Why not setting the allocation functions using the same utilities of `libuv`? Is there any problem with it?

hihig2001:
oh am I misunderstood something?

below codes are something I tracked.

```
template<typename T, typename U>
class StreamHandle: public Handle<T, U> {
    void read() {
        this->invoke(&uv_read_start, this->template get<uv_stream_t>(), &this->allocCallback, &readCallback);
    }
}
```
```
template<typename T, typename U>
class Handle: public Resource<T, U> {
    static void allocCallback(uv_handle_t *, std::size_t suggested, uv_buf_t *buf) {
        auto size = static_cast<unsigned int>(suggested);
        *buf = uv_buf_init(new char[size], size);
    }
}
```

I thought that problem rise this line of code

`*buf = uv_buf_init(new char[size], size);`

I'd like to pass pre-allocated buffer that I'm gonna re-use over and over.
In that way I understand that there should be a way that I can pass my pre-allocated buffer to uv_buf.base.

Is there are something that I misunderstood or miss some of functionality in libuv?

skypjack:
Hi, sorry for the late reply, I'm on vacation and also trying to have a _real_ break from my dev life. 😅 
I see what you mean. Consider that we're about to merge branch `v3` that is a **HUGE** change to the library, both in terms of performance and compilation time.
Can we consider this as a feature request to add allocator support to that version? It seems to me the only viable solution. Any other idea?

0EVSG:
+1 for the ability to inject a custom buffer allocation callback.

[...]

PR #290, opened 2023-06-08
Provide custom read buffer allocators for handle objects

MasterLogick:
I add one field `allocator` for handle objects, that is called as `uv_alloc_cb` in `uv_udp_recv_start` or `uv_recv_start`. Maybe allocator field should be private/protected or whatever. #289 #267

### DIFF
--- a/src/uvw/handle.hpp
+++ b/src/uvw/handle.hpp
@@ -243,6 +243,8 @@ public:
         uv_fileno(as_uv_handle(), &fd);
         return fd;
     }
+
+    std::function<void(uv_buf_t *, std::size_t, T &)> allocator = details::default_allocator<T>;
 };
 
 } // namespace uvw
--- a/test/uvw/tcp.cpp
+++ b/test/uvw/tcp.cpp
@@ -148,3 +148,56 @@ TEST(TCP, WriteError) {
     ASSERT_LT(handle->try_write(std::unique_ptr<char[]>{}, 0), 0);
     ASSERT_LT(handle->try_write(nullptr, 0), 0);
 }
+
+TEST(TCP, CustomAllocator) {
+    const std::string address = std::string{"127.0.0.1"};
+    const unsigned int port = 4242;
+
+    auto loop = uvw::loop::get_default();
+    auto server = loop->resource<uvw::tcp_handle>();
+    auto client = loop->resource<uvw::tcp_handle>();
+
+    auto *ptr = new char[5];
+
+    server->on<uvw::error_event>([](const auto &, auto &) { FAIL(); });
+    client->on<uvw::error_event>([](const auto &, auto &) { FAIL(); });
+
+    server->on<uvw::listen_event>([ptr](const uvw::listen_event &, uvw::tcp_handle &handle) {
+        std::shared_ptr<uvw::tcp_handle> socket = handle.parent().resource<uvw::tcp_handle>();
+
+        socket->on<uvw::error_event>([](const uvw::error_event &, uvw::tcp_handle &) { FAIL(); });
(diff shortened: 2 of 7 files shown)


src/uvw/handle.hpp
```diff
@@ -243,6 +243,8 @@ class handle: public resource<T, U, close_event, E...> {
         uv_fileno(as_uv_handle(), &fd);
         return fd;
     }
+
+    std::function<void(uv_buf_t *, std::size_t, T &)> allocator = details::default_allocator<T>;
```
skypjack, line 247: I'm definitely not a fan of public data members that expose implementation details. Like, at all.
Put aside that one can easily _reset_ the function and makes everything fail abruptly now. No good.

test/uvw/tcp.cpp
```diff
@@ -148,3 +148,56 @@ TEST(TCP, WriteError) {
...
+        socket->on<uvw::close_event>([&handle](const uvw::close_event &, uvw::tcp_handle &) { handle.close(); });
+        socket->on<uvw::end_event>([](const uvw::end_event &, uvw::tcp_handle &sock) { sock.close(); });
+        socket->on<uvw::data_event>([ptr](const uvw::data_event &evt, uvw::tcp_handle &sock) {
+            auto &d = const_cast<uvw::data_event &>(evt);
+            auto *data = d.data.release();
+            ASSERT_LE(evt.length, 5);
+            ASSERT_EQ(data, ptr);
+            sock.close();
+        });
+        socket->allocator = [ptr](uv_buf_t *b, size_t suggested, uvw::tcp_handle &h) {
```
skypjack, line 178: The idea was a little different. For example, since `uv_read_start` also accepts an allocator, `stream_handle<...>::read` can accept an allocator function as well (with a default) and forward it to the underlying function. Why the proposed solution should be better instead?

skypjack (changes requested):
The original proposal was way different than what I see here.
Moreover, the public data member is both risky and somewhat against the design of the whole library (that tries to encapuslate everything properly instead).
We can't really merge this as-is but I'm willing to review it again if you want to update it. Otherwise, I'll try to find some time during next week to implement the same feature as proposed initially. Just let me know. 👍

src/uvw/handle.hpp
MasterLogick: Ok, I admit it was a bad design.

test/uvw/tcp.cpp
MasterLogick: In my solution user can change allocator in event handlers. Also it allows to setup allocator once and do `start()`-`stop()` loops without passing custom function every time.

MasterLogick: Also forwarding directly to `uv_read_start` limits allocator implementation to real functions only. No lambda with capturing list

skypjack: Unfortunately, this is just an illusion if not even a mistake.

Imagine to set an allocator, then start a read request and finally change the allocator immediately after.
Because of how `libuv` works, the last allocator is invoked, although reading the code makes you think that the first one is used instead.
Now imagine to change the allocator _somewhere else_ in the codebase rather than _immediately after_. Maybe a couple of months after your change. Possibly a commit from a colleague or a PR from a random guy on the globe who doesn't know well how things work and what the expectation is.
I can already hear the curses of the dev who's called to fix this subtle bug. 😅 

When it comes to functional programming, passing something to the function that triggers the call is 99% of the time right or, well, at least predicatable, while _storing aside something for later uses_ gives a false feeling on what's going to happen and is more often than not a recipe for disasters.

MasterLogick: > although reading the code makes you think that the first one is used instead.

Well, for me it doesn't make.

I understand your points, but for me it is important to store alllocator. Will wait for your implementation.

Issue #294, opened 2023-08-15
suppress compiler warning


falsycat:
(sry for poor english)

I'm using uvw v3.2.0_libuv_v1.46 and want a way to suppress compiler warnings from header files of uvw because they're annoying.

I think it's the easiest solution to add `SYSTEM` keyword to `target_include_directories` in `CMakeLists.txt`.
I'll make PR if agreed.

I know the best one is fixing all warnings, however, doing that for all compilers will cost too much, I think.

thank you

falsycat:
@skypjack 

thank you for your reply!

warning:
```
/Users/falsycat/Repo/nf7/build/_deps/uvw-src/src/uvw/uv_type.hpp:19:34: error: unused parameter 'token' [-Werror=unused-parameter]
   19 |     explicit uv_type(loop::token token, std::shared_ptr<loop> ref) noexcept
      |                      ~~~~~~~~~~~~^~~~~
```

full compile command displayed by `make VERBOSE=1`:
```
cd /Users/falsycat/Repo/nf7/build/core && /opt/homebrew/bin/g++-13 -DUVW_AS_LIB -I/Users/falsycat/Repo/nf7/build -I/Users/falsycat/Repo/nf7 -I/Users/falsycat/Repo/nf7/build/_deps/uvw-src/src -I/Users/falsycat/Repo/nf7/build/_deps/libuv-src/include -isystem /Users/falsycat/Repo/nf7/build/_deps/luajit-src/src -g -std=gnu++23 -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk -mmacosx-version-min=11.0 -Wall -Wextra -Wpedantic -Werror -MD -MT core/CMakeFiles/nf7_core.dir/uv/concurrency.cc.o -MF CMakeFiles/nf7_core.dir/uv/concurrency.cc.o.d -o CMakeFiles/nf7_core.dir/uv/concurrency.cc.o -c /Users/falsycat/Repo/nf7/core/uv/concurrency.cc
```

CMake script about uvw:
```
FetchContent_Declare(
  uvw
  GIT_REPOSITORY https://github.com/skypjack/uvw.git
  GIT_TAG        v3.2.0_libuv_v1.46
)
set(BUILD_UVW_LIBS ON)
FetchContent_MakeAvailable(uvw)
```

compiler version:
```
$ g++-13 --version
g++-13 (Homebrew GCC 13.1.0) 13.1.0
[...]

PR #296, opened 2023-08-21
uv_type: supress `-Werror=unused` compiler error

aloisklink:
The `uv_type()` constructor does not use its `token` parameter, which causes a `-Werror=unused` compiler error on GCC (and probably on other compilers too).

~Luckily, C++17 adds the [`[[maybe_unused]]`][1] attribute that all standards compliant compilers will support to hide unused errors.~ Removed, see https://github.com/skypjack/uvw/pull/296#issuecomment-1686169140.

[1]: https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

---

This fixes the error reported in https://github.com/skypjack/uvw/issues/294#issuecomment-1683625763, but I didn't want to add a `Fixes: #294` into my commit or PR, since it seems like that issue is more about disabling all potential warnings using CMake, not just this single error.

### DIFF
--- a/src/uvw/uv_type.hpp
+++ b/src/uvw/uv_type.hpp
@@ -16,7 +16,7 @@ namespace uvw {
  */
 template<typename U>
 struct uv_type {
-    explicit uv_type(loop::token token, std::shared_ptr<loop> ref) noexcept
+    explicit uv_type(loop::token, std::shared_ptr<loop> ref) noexcept
         : owner{std::move(ref)}, resource{} {}
 
     uv_type(const uv_type &) = delete;


skypjack:
Oh, well, since the parameter name is the same of the type, I'd just drop the first one. As in:
```cpp
explicit uv_type(loop::token, std::shared_ptr<loop> ref) noexcept
```
Rather than:
```cpp
explicit uv_type([[maybe_unused]] loop::token token, std::shared_ptr<loop> ref) noexcept
```

aloisklink:
> Oh, well, since the parameter name is the same of the type, I'd just drop the first one.

That makes sense! You only really need `[[maybe_unused]]` if you need to give the parameter a name (e.g. for Doxygen). I've fixed this in https://github.com/skypjack/uvw/pull/296/commits/619bf5d9a0fa6cfa3174ef1ec879e2790b13c5a9, and I've updated the commit/PR description to remove any references to `[[maybe_unused]]`.

skypjack:
Thank you (and sorry for the late reply, I'm out of home).

skypjack (approved):

aloisklink:
> Thank you (and sorry for the late reply, I'm out of home).

No worries! 2 days isn't really a late reply! I might take a while to reply too, since although we're generally allowed to submit patches to open-source, we're only allowed to do it in our personal time!

PR #302, opened 2023-08-23
Fix compiler warnings in `test` directory

aloisklink:
Fixes a couple of compiler warnings in the test directory.

Since the `test` directory doesn't affect users, I've just made a single PR that includes all my changes. But please let me know if you'd prefer to break my changes up into smaller PRs (I've tried to split up each change into individual commits, to make it a bit easier to review!)

### DIFF
--- a/test/main.cpp
+++ b/test/main.cpp
@@ -65,7 +65,7 @@ void conn(uvw::loop &loop) {
 
         auto dataTryWrite = std::unique_ptr<char[]>(new char[1]{'a'});
         int bw = handle.try_write(std::move(dataTryWrite), 1);
-        std::cout << "written: " << ((int)bw) << std::endl;
+        std::cout << "written: " << static_cast<int>(bw) << std::endl;
 
         auto dataWrite = std::unique_ptr<char[]>(new char[2]{'b', 'c'});
         handle.write(std::move(dataWrite), 2);
--- a/test/main.cpp
+++ b/test/main.cpp
@@ -28,7 +28,7 @@ void listen(uvw::loop &loop) {
         std::cout << "remote: " << remote.ip << " " << remote.port << std::endl;
 
         client->on<uvw::data_event>([](const uvw::data_event &event, uvw::tcp_handle &) {
-            std::cout.write(event.data.get(), event.length) << std::endl;
+            std::cout.write(event.data.get(), static_cast<std::streamsize>(event.length)) << std::endl;
             std::cout << "data length: " << event.length << std::endl;
         });
 
--- a/test/uvw/fs_req.cpp
+++ b/test/uvw/fs_req.cpp
@@ -107,7 +107,7 @@ TEST(FsReq, Stat) {
     fsReq->on<uvw::error_event>([](const auto &, auto &) { FAIL(); });
     fileReq->on<uvw::error_event>([](const auto &, auto &) { FAIL(); });
 
-    fsReq->on<uvw::fs_event>([&](const auto &event, auto &req) {
+    fsReq->on<uvw::fs_event>([&](const auto &event, auto &) {
         if(event.type == uvw::fs_req::fs_type::STAT) {
             ASSERT_FALSE(checkFsStatEvent);
             checkFsStatEvent = true;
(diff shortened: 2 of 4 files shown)


test/uvw/stream.cpp
```diff
@@ -12,4 +12,8 @@ struct fake_stream_handle: uvw::stream_handle<fake_stream_handle, fake_stream_t>
     int init(Args &&...) {
         return 0;
     }
+
+    virtual int init() override {
```
skypjack, line 16: Sorry for the slow reviews but I'm out of home for another couple of days.
Can I ask you why did you add this function? I don't see the purpose at first glance. Thanks.

skypjack (commented):
LGTM but I don't get why the `init` function was added. Any hints?

test/uvw/stream.cpp
aloisklink: It prevents the following warning:

```
src/uvw/uv_type.hpp:32:17: warning: ‘int uvw::uv_type<U>::init() [with U = fake_stream_t]’ was hidden [-Woverloaded-virtual]
   32 |     virtual int init() {
      |                 ^~~~
test/uvw/stream.cpp:12:9: note:   by ‘fake_stream_handle::init(Args&& ...)’
   12 |     int init(Args &&...) override {
                ^~~~
```

I've added some basic details in the commit description: https://github.com/skypjack/uvw/pull/302/commits/628b60339559339f6bb4089199f3d6af05841149.

Some details on why this is a problem is here: https://isocpp.org/wiki/faq/strange-inheritance#hiding-rule

But basically, the `template<typename... Args> int init(Args &&...) { return 0;}` function **doesn't** override the `virtual init();` function in https://github.com/skypjack/uvw/blob/113772521d65501914a1bcd769437aac28b10307/src/uvw/uv_type.hpp#L28-L34 because it's a `template` function (however, it does hide the original function). If you want to override it, we need to explicitly declare a non-`template` version of it.

Alternatively, we could also add a `using stream_handle::init;` statement, and that should also fix the error, but I'm guessing you're overriding `init()` for a reason here.

---

> Sorry for the slow reviews

And no worries about the slow reviews :) I help maintain the https://github.com/mermaid-js/mermaid project, and it sometimes takes me weeks to find the time to review really big PRs!

skypjack: I really, *REALLY*, really feel like `init` doesn't have to be `virtual` not in the public interface of the base class actually.
Let me try. If this is the case, I'll push a commit and close this PR without merging it. Gimme a minute. 👍

aloisklink: Awesome! https://github.com/skypjack/uvw/commit/dea293ffaa5ee68c0152531e03453935a44bbfcb seems to have fixed this warning!

By the way, what should we do with the other commits in this PR? Would you like me to make separate PRs for each commit?

Alternatively, we could re-open this PR, and we could just drop commit https://github.com/skypjack/uvw/pull/302/commits/628b60339559339f6bb4089199f3d6af05841149 if you'd prefer to have one big PR fixing all the warnings in the `test/` directory, as although each commit in this PR fixes a separate warning, the `test/` directory isn't as important as anything in the `src/` directory.

skypjack: Yeah, sorry. I'd reopen this one and strip the commit for the `init` function.
Can you do that? Otherwise I can port everything to a new branch and merge it directly.
Just let me know. Both are fine for me and sorry again for the mistake on closing this one. 🙏

aloisklink: > I'd reopen this one and strip the commit for the init function.

Done! It was even the last commit, so all the other commits have the same hash.

> Otherwise I can port everything to a new branch and merge it directly.

It's up to you, it might be better to do a **Rebase and merge** instead of a **Squash and merge**, since each commit is a logically separate change, but it's your GitHub repo, so you can do what you prefer!

> Both are fine for me and sorry again for the mistake on closing this one. 🙏

And no worries! I've done it a couple times where I wrote something like, `This doesn't fix #xxxx`, but GitHub sees `fix #xxxx` and auto-closes the issue.

aloisklink:
I've dropped commit https://github.com/skypjack/uvw/commit/628b60339559339f6bb4089199f3d6af05841149 as discussed in https://github.com/skypjack/uvw/pull/302#discussion_r1311405389, since https://github.com/skypjack/uvw/commit/dea293ffaa5ee68c0152531e03453935a44bbfcb has already fixed this issue.

I'm not 100% sure why CI is failing, but I'm guessing it might just be a random failure, since everything seems to be working on my `aloisklink/uvw` fork: https://github.com/aloisklink/uvw/commit/b820447a60d49a8a3c8503930366918f76515089. Can you re-run GitHub Actions and see if that fixes it?

skypjack (approved):

PR #304, opened 2023-08-23
Check for GCC/Clang compiler warnings in Ubuntu GitHub Actions CI

aloisklink:
Add a [`CMakePresets.json`][1] file that contains `ci-ubuntu` preset. This preset contains most of the recommended compiler warnings from the [`cmake-init`][2] project, with the following changes:
  - I had to remove the `-Wdouble-promotion` and `-Werror=float-equal` flags since there are issues within the GTest headers.
  - For compatibility with g++-7, which doesn't support this flag, `-Wextra-semi` was removed.
  - I also had to remove some of the following flags like: `-fstack-protector-strong -fstack-clash-protection`. These flags cause errors in Clang v10, and to be honest, with stuff like address sanitizer, we probably don't need them.

I've also set the `"CMAKE_CXX_FLAGS_DEBUG": "-Werror"` variable when running on `ubuntu` in GitHub Actions. This will make the `ubuntu` build fail whenever any warnings are seen in the code, so you'll instantly spot it whenever somebody tries to add code that has warnings to this project. This should prevent issues like https://github.com/skypjack/uvw/issues/294 from happening in the future.

[1]: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html
[2]: https://github.com/friendlyanon/cmake-init/blob/aa42211c79ab5117b05a2d9f795427f078a0a3d5/cmake-init/templates/common/CMakePresets.json#L88

---

~I've left this PR as a draft, since the CI currently fails, but after the following PRs are merged, this CI job should work~ All of these PRs have now been merged, so CI passes:
  - https://github.com/skypjack/uvw/pull/296
  - https://github.com/skypjack/uvw/pull/297
  - https://github.com/skypjack/uvw/pull/298
  - https://github.com/skypjack/uvw/pull/299
  - https://github.com/skypjack/uvw/pull/300
  - https://github.com/skypjack/uvw/pull/301
  - https://github.com/skypjack/uvw/pull/302
  - https://github.com/skypjack/uvw/pull/303

### DIFF
--- a/.github/workflows/build-ubuntu-20.04.yml
+++ b/.github/workflows/build-ubuntu-20.04.yml
@@ -27,12 +27,11 @@ jobs:
         sudo apt-get update --fix-missing
         sudo apt install -y ${{ matrix.compiler.pkg }}-${{ matrix.compiler.version }}
     - name: Compile tests
-      working-directory: build
       env:
         CXX: ${{ matrix.compiler.exe }}-${{ matrix.compiler.version }}
       run: |
-        cmake ${{ matrix.mode }} -DBUILD_TESTING=ON -Dlibuv_buildtests=OFF ..
-        make -j2
+        cmake ${{ matrix.mode }} --preset ci-ubuntu
+        cmake --build build/ --parallel 2
     - name: Run tests
       working-directory: build
       env:
--- a/.github/workflows/build-ubuntu-latest.yml
+++ b/.github/workflows/build-ubuntu-latest.yml
@@ -26,12 +26,11 @@ jobs:
         sudo apt-get update --fix-missing
         sudo apt install -y ${{ matrix.compiler.pkg }}-${{ matrix.compiler.version }}
     - name: Compile tests
-      working-directory: build
       env:
         CXX: ${{ matrix.compiler.exe }}-${{ matrix.compiler.version }}
       run: |
-        cmake ${{ matrix.mode }} -DBUILD_TESTING=ON -Dlibuv_buildtests=OFF ..
-        make -j2
+        cmake ${{ matrix.mode }} --preset ci-ubuntu
+        cmake --build build/ --parallel 2
     - name: Run tests
       working-directory: build
       env:
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,59 @@
+{
+  "version": 2,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 13,
+    "patch": 0
+  },
+  "configurePresets": [
+    {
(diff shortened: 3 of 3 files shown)


skypjack:
`-Dlibuv_buildtests=OFF` is gone during the rework. Am I wrong?

aloisklink:
> `-Dlibuv_buildtests=OFF` is gone during the rework. Am I wrong?

I've added it to the `dev-mode` CMake preset:

https://github.com/skypjack/uvw/blob/8a23e5808635b17e80f887482ec5f5f8cf33d3c6/CMakePresets.json#L26-L29

It will make it easier for us developers to run `cmake --preset dev-mode` locally, and in the future, we can add a `ci-windows`/`ci-darwin` presets for doing the Windows/MacOS CI runs, that also inherit from the `dev-mode` preset, so we'll only have to modify everything in a single file.

The only problem is that is [`cmake-presets` are only supported in CMake 3.19 or later](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html), but since we're only using them for CI/development (they won't be used by end-users), this isn't too big of an issue for us!

aloisklink:
I've `git rebase`-d onto https://github.com/skypjack/uvw/commit/683883b08a6fdf19edc48d99ac1b901924a574cd on the `main` branch, since all the PRs have been merged and all the GCC/Clang compiler warnings have been fixed!

skypjack:
Before merging, is this the last PR in the row or is there something else to address?
Doing this work on warnings over the holidays made it a bit tricky to follow, sorry. 🙂

skypjack (approved):

aloisklink:
> Before merging, is this the last PR in the row or is there something else to address?

This is the last PR in the row. Every other PR I made was about fixing warnings. This PR is just here to make big red cross in GitHub Actions if any new code is added that has warnings in it.

I considered doing something similar for Windows/MacOS, but I don't have easy access to a Windows/MacOS machine, so I guess that can wait until a Windows/MacOS user complains about warnings.

I do also want to make a PR fixing the https://skypjack.github.io/uvw/ website (currently a lot of links don't work, since they point to https://skypjack.github.io/uvw/classuvw_1_1emitter.html, but the file is actually called https://skypjack.github.io/uvw/classuvw_1_1Emitter.html (e.g. with a capital `E` not `e`)), but that's completely unrelated to this compiler warnings series of PRs.

> Doing this work on warnings over the holidays made it a bit tricky to follow, sorry. 🙂

And no worries :) Thanks for reviewing stuff on your holiday! To be honest, I wouldn't have minded waiting. After all, since these PRs were mostly just refactoring PRs, there's probably no point in making a `uvw` release just for this :smile:

PR #317, opened 2024-10-16
fix: fixed incorrect wrap for process's stdio container

sssooonnnggg:
The stdio api of the process should ensure the calling order.

According to libuv's documentation and test cases, where stdio[0] is designated as stdin, stdio[1] as stdout, and stdio[2] as stderr. 

When specifying stdin for the child process as stdin_pipe, stdout to a custom file, and stderr as stderr_pipe:

process_handle.stdio(stdin_pipe, flag)
process_handle.stdio(stdout_file_handle, flag)
process_handle.stdio(stderr_pipe, flag)

The previous uvw implementation incorrectly concatenates it as follows because it used two vector:

stdio[0] = stdout_file_handle
stdio[1] = stdin_pipe
stdio[2] = stderr_pipe

This is not the correct behaviour

This PR try to fix this.😊

Libuv spawn examples: https://github.com/libuv/libuv/blob/v1.x/test/test-spawn.c

### DIFF
--- a/src/uvw/process.cpp
+++ b/src/uvw/process.cpp
@@ -44,13 +44,8 @@ UVW_INLINE int process_handle::spawn(const char *file, char **args, char **env) 
     po.uid = po_uid;
     po.gid = po_gid;
 
-    std::vector<uv_stdio_container_t> poStdio;
-    poStdio.reserve(po_fd_stdio.size() + po_stream_stdio.size());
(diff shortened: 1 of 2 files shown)


sssooonnnggg:
Here we can see the convention of how libuv deal with the indexing of stdio，0 is stdin，1 is stdout，2 is stderr
https://github.com/libuv/libuv/blob/fbe2d85bd5a5c370a8cacea92b3bdfbd9f98a530/src/win/process.c#L1018-L1020

PR #319, opened 2024-11-27
Initialize some uninitialized variables

fedegiova:
Found some class variables that were not initialized so they contained garbage values.


fedegiova:
Done

PR #327, opened 2025-11-25
fs_event: enable bit operations for the uvw_fs_event enum

fcelda:
This PR adds the _UVW_ENUM_ member to the enum class _uvw_fs_event_ to enables the bitwise operations over the type.

The libuv documentation says that the flags are a bit mask:

> The events parameter is an ORed mask of [uv_fs_event](https://docs.libuv.org/en/v1.x/fs_event.html#c.uv_fs_event) elements.

-- https://docs.libuv.org/en/v1.x/fs_event.html#c.uv_fs_event_cb:


skypjack:
I fixed the issue with the thread class. Can you rebase on `main` and rerun the tests for a double check? Thanks. 🙏

PR #332, opened 2026-03-22
Add include directories to vendored uv target

russkel:
Hi,

I am using uvw and find the vendored uv static lib handy. However I was not able to simply just link to uvw and uv as it would result in missing include issues, because the uv directory is put into a nested include directory.

I am not sure if this was intentional or a mistake - if it was me I would have uv and uvw header directories at the same level and not nested. In any case, this fixes the missing include error.

In addition, any of the `uv::` namespace does not seem to be exported either as they are build time aliases. I changed the README to reflect that.

I bumped to latest libuv because I needed TCP keepalive_ex functionality. Let me know if you want me to drop that commit.


russkel:
@skypjack ping

